PR #53, opened 2020-09-25
Add Record.Copy

arl:
#### :question: What

Add `Copy` method to the `Record` interface.
This is necessary for filters that want to create copies of a record.
Without that new method, one would have to call `src.ToText` followed by `dst.Parse` to make `dst` being a copy of `src` `Record`. This is what I call the naive solution.

The following benchmarks comparisons shows the optimized Copy vs the naive version:

    name                  old time/op    new time/op    delta
    LogLineCopy/set=0-8     7.94µs ± 8%    4.65µs ±10%  -41.43%  (p=0.000 n=15+15)
    LogLineCopy/set=1-8     32.7µs ± 5%    30.8µs ± 6%   -5.86%  (p=0.000 n=13+15)
    LogLineCopy/set=10-8    33.8µs ± 3%    30.6µs ± 5%   -9.56%  (p=0.000 n=14+15)
    LogLineCopy/set=20-8    32.8µs ± 6%    30.1µs ± 4%   -8.03%  (p=0.000 n=15+15)

    name                  old alloc/op   new alloc/op   delta
    LogLineCopy/set=0-8     24.5kB ± 0%    24.5kB ± 0%     ~     (all equal)
    LogLineCopy/set=1-8     39.0kB ± 0%    31.3kB ± 0%  -19.71%  (p=0.000 n=15+15)
    LogLineCopy/set=10-8    39.0kB ± 0%    31.3kB ± 0%  -19.71%  (p=0.000 n=15+15)
    LogLineCopy/set=20-8    39.3kB ± 0%    31.3kB ± 0%  -20.50%  (p=0.000 n=15+15)

    name                  old allocs/op  new allocs/op  delta
    LogLineCopy/set=0-8       3.00 ± 0%      3.00 ± 0%     ~     (all equal)
    LogLineCopy/set=1-8       6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.000 n=15+15)
    LogLineCopy/set=10-8      6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.000 n=15+15)
    LogLineCopy/set=20-8      6.00 ± 0%      4.00 ± 0%  -33.33%  (p=0.000 n=15+15)

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/logline.go
+++ b/logline.go
@@ -113,7 +113,7 @@ var errLogLineTooManyFields = errors.New("LogLine has too many fields")
 // instance. For performance reasons, it doesn't reset all the writable fields
 // of the line. If you want to use Parse over an already parsed LogLine, use
 // Clear before.
-func (l *LogLine) Parse(text []byte, meta *Metadata) error {
+func (l *LogLine) Parse(text []byte, meta Metadata) error {
 	l.idx[0] = -1
 	fc := FieldIndex(1)
 	for i, ch := range text {
@@ -130,7 +130,7 @@ func (l *LogLine) Parse(text []byte, meta *Metadata) error {
 	}
 	l.data = text
 	if meta != nil {
-		l.meta = *meta
+		l.meta = meta
 	}
 
 	return nil
--- a/logline.go
+++ b/logline.go
@@ -208,13 +208,39 @@ func (l *LogLine) Cache() *Cache {
 
 // Copy creates and returns a copy of the current log line.
 func (l *LogLine) Copy() Record {
+	// Copy metadata
 	md := make(Metadata)
-
 	for k, v := range l.meta {
 		md[k] = v
 	}
 
-	cpy := LogLine{}
-	cpy.Parse(l.ToText(nil), md)
-	return &cpy
+	cpy := &LogLine{
+		cache:          l.cache,
+		wcnt:           l.wcnt,
+		meta:           md,
+		FieldSeparator: l.FieldSeparator,
+	}
+
+	if l.data != nil {
+		cpy.data = make([]byte, len(l.data))
+		copy(cpy.data, l.data)
+	}
+
+	if cpy.wcnt == 0 {
+		// Log line hasn't been modified so we can early exit and
+		// let idx, wmask and wdata to their zero-values.
+		return cpy
+	}
+
+	copy(cpy.idx[:], l.idx[:])
+	copy(cpy.wmask[:], l.wmask[:])
+
+	for i := 0; i < len(l.wdata); i++ {
+		if l.wdata[i] != nil {
+			cpy.wdata[i] = make([]byte, len(l.wdata[i]))
+			copy(cpy.wdata[i], l.wdata[i])
+		}
+	}
+
+	return cpy
 }
--- a/logline.go
+++ b/logline.go
@@ -205,3 +205,16 @@ func (l *LogLine) Meta(key string) (interface{}, bool) {
 func (l *LogLine) Cache() *Cache {
 	return &l.cache
 }
+
+// Copy creates and returns a copy of the current log line.
+func (l *LogLine) Copy() Record {
+	md := make(Metadata)
+
+	for k, v := range l.meta {
+		md[k] = v
+	}
+
+	cpy := LogLine{}
+	cpy.Parse(l.ToText(nil), md)
+	return &cpy
+}
--- a/logline.go
+++ b/logline.go
@@ -232,14 +232,14 @@ func (l *LogLine) Copy() Record {
 		return cpy
 	}
 
-	copy(cpy.idx[:], l.idx[:])
-	copy(cpy.wmask[:], l.wmask[:])
+	// Copy read-only fields
+	cpy.idx = l.idx
 
-	for i := 0; i < len(l.wdata); i++ {
-		if l.wdata[i] != nil {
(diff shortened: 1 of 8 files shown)


logline.go
```diff
@@ -205,3 +205,38 @@ func (l *LogLine) Meta(key string) (interface{}, bool) {
...
+		wcnt:           l.wcnt,
+		meta:           md,
+		FieldSeparator: l.FieldSeparator,
+	}
+
+	if cpy.wcnt != 0 {
+		// If the log line has been modified, benchmarks have proven
+		// that it's more efficient to serialize and reparse to perform
+		// a copy (both in terms of time and allocation.
+		cpylen := len(l.data) + len(l.data)/2
```
tommyblue, line 228: please add a brief comment to explain why you choose to use this length

tommyblue: and, as you said, the `ToText` fn already allocates the right amount of memory when it calculates `wlen` so you can probably avoid using `cpylen`

logline.go
```diff
@@ -205,3 +205,38 @@ func (l *LogLine) Meta(key string) (interface{}, bool) {
...
+		text := l.ToText(make([]byte, 0, cpylen))
+		cpy.Parse(text, md)
```
tommyblue, line 230: uhm, but at this point `cpy.wcnt` isn't wrong?

tommyblue: if I'm not wrong after a parse it should be `0` but it's now `l.wcnt`

arl: You're right it should be 0 since all fields have been parsed and no fields have been written yet.
The line above where I set wcnt at LogLine creation is a leftover of a previous implementation and should be removed.

I'm surprised the tests didn't catch that though. I'm gonna write one that does.

logline.go
arl: Yes, It's probably safe to keep the same length as org.data and let ToText takes care of the actual length

arl: ToText only estimates the actual length in order to reduce the number of allocations, here is the results of various benchmarks looking at various size for the pre-allocated buffer, for example `bench.len+50` means the benchmark where the preallocated buffer is the original + length + 50% (to account for written fields).
bench.len+50 (current code) is the best all-round (cpu, allocations bytes + count)

```
name \ time/op        bench.len+50  bench.len+25  bench.len+10  bench.len    bench.nil
LogLineCopy/set=0-8    4.73µs ± 5%   5.03µs ± 2%   4.62µs ± 8%  4.74µs ± 2%  4.67µs ± 3%
LogLineCopy/set=1-8    30.4µs ± 8%   31.7µs ± 6%   31.2µs ± 2%  32.6µs ± 5%  31.1µs ± 0%
LogLineCopy/set=10-8   30.3µs ± 4%   31.0µs ± 4%   32.6µs ± 1%  32.7µs ± 7%  32.4µs ± 2%
LogLineCopy/set=20-8   29.7µs ± 7%   29.6µs ± 2%   31.0µs ± 4%  31.8µs ± 5%  30.5µs ± 4%

name \ alloc/op       bench.len+50  bench.len+25  bench.len+10  bench.len    bench.nil
LogLineCopy/set=0-8    24.5kB ± 0%   24.5kB ± 0%   24.5kB ± 0%  24.5kB ± 0%  24.5kB ± 0%
LogLineCopy/set=1-8    31.3kB ± 0%   34.7kB ± 0%   34.4kB ± 0%  41.6kB ± 0%  39.0kB ± 0%
LogLineCopy/set=10-8   31.3kB ± 0%   34.7kB ± 0%   34.4kB ± 0%  41.6kB ± 0%  39.0kB ± 0%
LogLineCopy/set=20-8   31.3kB ± 0%   34.7kB ± 0%   34.4kB ± 0%  42.0kB ± 0%  39.3kB ± 0%

name \ allocs/op      bench.len+50  bench.len+25  bench.len+10  bench.len    bench.nil
LogLineCopy/set=0-8      3.00 ± 0%     3.00 ± 0%     3.00 ± 0%    3.00 ± 0%    3.00 ± 0%
LogLineCopy/set=1-8      4.00 ± 0%     5.00 ± 0%     5.00 ± 0%    7.00 ± 0%    6.00 ± 0%
LogLineCopy/set=10-8     4.00 ± 0%     5.00 ± 0%     5.00 ± 0%    7.00 ± 0%    6.00 ± 0%
LogLineCopy/set=20-8     4.00 ± 0%     5.00 ± 0%     5.00 ± 0%    7.00 ± 0%    6.00 ± 0%
```

logline.go
```diff
@@ -205,3 +205,38 @@ func (l *LogLine) Meta(key string) (interface{}, bool) {
...
+
+		return cpy
+	}
+
+	// If the log line hasn't been modified it's more efficient to recreate it
+	// from scratch and copying data (log line internal buffer).
+	if l.data != nil {
+		cpy.wcnt = l.wcnt
```
tommyblue: I think this is useless, `l.wcnt` is zero as well as `cpy.wcnt`

tommyblue (approved):
LGTM, just check the comment

logline.go
tommyblue: ok, I see you removed it 👍

arl: Yeah sorry😂

logline.go
arl: I still need to add the comment explaining why len+50%

PR #56, opened 2020-09-30
Add sqlite output

tommyblue:
#### :question: What

Add the SQLite output

### DIFF
--- a/output/sqlite.go
+++ b/output/sqlite.go
@@ -1,0 +1,412 @@
+package output
+
+import (
+	"bytes"
+	"database/sql"
+	"fmt"
+	"html/template"
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	"github.com/AdRoll/baker"
+)
+
+// SQLiteDesc declares the standard (non-raw) SQLite output.
+var SQLiteDesc = baker.OutputDesc{
+	Name:   "SQLite",
+	New:    newSQLiteWriter(false),
+	Config: &SQLiteWriterConfig{},
+	Help:   "Writes a chosen set of fields as table columns into a local SQLite database file",
+	Raw:    false,
+}
+
+// SQLiteRawDesc declares the raw SQLite output.
+var SQLiteRawDesc = baker.OutputDesc{
+	Name:   "SQLiteRaw",
+	New:    newSQLiteWriter(true),
+	Config: &SQLiteRawWriterConfig{},
+	Help:   "Writes a chosen set of fields, plus the raw record, as table columns into a local SQLite database file",
+	Raw:    true,
+}
+
+// SQLiteWriterConfig holds the configuration parameters for the
+// standard SQLite baker output.
+type SQLiteWriterConfig struct {
+	PathString string   `help:"Path to local SQLite file to write the results to. Will be created if it does not exist. Can contain {{.ShardId}} and {{.Field}} for replacement" required:"true"`
+	TableName  string   `help:"Table name to which to write the records to." required:"true"`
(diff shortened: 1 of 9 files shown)


output/all.go
```diff
@@ -5,11 +5,15 @@ import (
...
+
+func init() {
+	All = append(All, []baker.OutputDesc{
+		DynamoDBDesc,
+		FileWriterDesc,
+		NopDesc,
+		OpLogDesc,
+		StatsDesc,
+		WebSocketDesc,
+	}...)
```
arl, line 18: I don't think this should be changed, first because it's always preferable to not use `init` if possible, second because the result is the same as the previous code

output/sqlite.go
```diff
@@ -0,0 +1,422 @@
+// +build cgo_sqlite
```
arl, line 1: I know at the moment we don't give the possibility to use another sqlite driver.
But... in theory the sqlite baker output shouldn't care about which particular sqlite driver is used under the hood so I tihnk it would be best that cgo does not appear here.

Generally for this kind of stuff, we'd have something along the lines of:


 - sqlite_cgo.go:

```go
// +build cgo_sqlite
package output

import (_ "github.com/mattn/go-sqlite3")
```

- sqlite_go.go:
```go
// +build !cgo_sqlite
package output

import (_ "/whatever/pure-go/package")
```

If we do this, we'll always have one, but only one, sqlite driver that is loaded, so we don't need to conditionally add `SQLiteDesc` to the `output.All`. 

Again, even if we don't have a pure-go package at the moment, we should still do this, but of course omit loading the package in `sqlite_go.go`. That means that if somebody tries to compile without providing `cgo_sqlite` it will just work, until they try to use the sqlite output, it will fail at startup saying it can't find an existing sqlite driver (which is exactly what we want)

output/sqlite.go
```diff
@@ -0,0 +1,422 @@
...
+
+	_ "github.com/mattn/go-sqlite3"
+
+	"github.com/AdRoll/baker"
+)
+
+func init() {
+	All = append(All, []baker.OutputDesc{SQLiteDesc, SQLiteRawDesc}...)
+	sort.Slice(All, func(i, j int) bool { return All[i].Name < All[j].Name })
+}
```
arl, line 23: same as before, `init` is not needed.
But even if that was the case, I would let whoever needs `All` sorted the responsibility to sort it (in our case the help renderer)

output/sqlite.go
```diff
@@ -0,0 +1,422 @@
...
+	"bytes"
+	"database/sql"
+	"fmt"
+	"html/template"
+	"path/filepath"
+	"sort"
+	"strings"
+	"unicode"
+
+	_ "github.com/mattn/go-sqlite3"
```
arl, line 15: see above comments

arl (commented):
requesting changes about the handling of the build tag and `init`.

output/all.go
tommyblue: ok

output/sqlite_cgo.go
```diff
@@ -0,0 +1,13 @@
...
+package output
+
+import (
+	_ "github.com/mattn/go-sqlite3"
+
+	"github.com/AdRoll/baker"
+)
+
+func init() {
+	All = append(All, []baker.OutputDesc{SQLiteDesc, SQLiteRawDesc}...)
```
arl, line 12: This works:
```suggestion
	All = append(All, SQLiteDesc, SQLiteRawDesc)
```

arl (approved):
LGTM

there's a single code simplification comment

Should we modify the github actions script to check that we build both with and without cgo_sqlite?

PR #58, opened 2020-10-13
Improve API documentation

tommyblue:
#### :question: What

Improve API docs for https://pkg.go.dev/github.com/AdRoll/baker

### DIFF
--- a/api.go
+++ b/api.go
@@ -3,8 +3,8 @@ package baker
 // Data represents raw data consumed by a baker input, possibly
 // containing multiple records before they're parsed.
 type Data struct {
-	Bytes []byte   // The raw content
-	Meta  Metadata // Some metadata, filled in by the input
+	Bytes []byte   // Bytes is the slice of raw bytes read by an input
+	Meta  Metadata // Meta is filled by the input and holds metadata that will be associated to the records parsed from Bytes
 }
 
 // Metadata about the input data; each Input will directly populate this
@@ -21,7 +21,7 @@ func (m *Metadata) get(key string) (val interface{}, ok bool) {
 }
 
 // InputStats contains statistics about the input component,
-// ready for export to the metric client and to print debug info
+// ready for export to the metric client and to print debug info.
 type InputStats struct {
 	NumProcessedLines int64
 	CustomStats       map[string]string
--- a/api.go
+++ b/api.go
@@ -1,7 +1,7 @@
 package baker
 
-// Data represents a bunch of bytes retrieved from the input,
-// before parsing them to a Record
+// Data represents raw data consumed by a baker input, possibly
+// containing multiple records before they're parsed.
 type Data struct {
 	Bytes []byte   // The raw content
 	Meta  Metadata // Some metadata, filled in by the input
--- a/api.go
+++ b/api.go
@@ -1,14 +1,10 @@
 package baker
 
-type InputStats struct {
-	NumProcessedLines int64
-	CustomStats       map[string]string
-	Metrics           MetricsBag
-}
-
+// Data represents a bunch of bytes retrieved from the input,
+// before parsing them to a Record
 type Data struct {
-	Bytes []byte
-	Meta  Metadata
+	Bytes []byte   // The raw content
+	Meta  Metadata // Some metadata, filled in by the input
 }
 
 // Metadata about the input data; each Input will directly populate this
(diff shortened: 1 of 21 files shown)


desc.go
```diff
@@ -73,6 +76,9 @@ type InputDesc struct {
 }
 
 // FilterDesc describes a Filter component to the topology.
+//
+// It has a name, a config object, a constructor function (New)
+// and a help string
```
arl, line 81: same

logline.go
```diff
@@ -192,6 +191,7 @@ func (l *LogLine) ToText(buf []byte) []byte {
 	return buf
 }
 
+// Clear the logline, overwriting with a new logline, only inheriting the field separator
```
arl, line 194:
```suggestion
// Clear clears the logline...
```

The fact that it's overwritten with a new logline is an implementation detail.

FYI: On `Record.Clear` the comment is 
> // Clear clears the record internal state, making it empty.

arl (commented):
Some comments about conventions, some propositions and obviously 1 or 2 nitpicks, just because :D.

api.go
```diff
@@ -1,14 +1,10 @@
 package baker
 
-type InputStats struct {
-	NumProcessedLines int64
-	CustomStats       map[string]string
-	Metrics           MetricsBag
-}
-
+// Data represents a bunch of bytes retrieved from the input,
+// before parsing them to a Record
```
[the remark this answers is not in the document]
tommyblue:
```suggestion
// Data represents raw data consumed by a baker input, possibly
// containing multiple records before they're parsed.
```

config.go
```diff
@@ -44,6 +44,12 @@ type ConfigInput struct {
 
 // ConfigFilterChain specifies the configuration for the whole fitler chain.
 type ConfigFilterChain struct {
+	// Procs defines the number of parallel filtering processes. As a single records
```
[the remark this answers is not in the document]
tommyblue: 💘

desc.go
```diff
@@ -3,20 +3,20 @@ package baker
...
-	Validate      ValidationFunc
-	CreateRecord  func() Record
+	ShardingFuncs map[FieldIndex]ShardingFunc // functions to calculate sharding based on field index
+	Validate      ValidationFunc              // function to use to validate a Records
+	CreateRecord  func() Record               // create a new record
 
-	FieldByName func(string) (FieldIndex, bool)
-	FieldName   func(FieldIndex) string
+	FieldByName func(string) (FieldIndex, bool) // get a field index by its name
+	FieldName   func(FieldIndex) string         // gets a field name by its index
```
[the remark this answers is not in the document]
tommyblue: 🤦

baker.go
```diff
@@ -1,17 +1,30 @@
+/*
+Package baker can be used as library to build a baker binary (see the examples/ folder).
```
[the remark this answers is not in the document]
tommyblue: I made an attempt. As we have a complex readme, I didn't add much to that docstring

baker.go
```diff
@@ -1,17 +1,28 @@
+/*
+Package baker provides types and functions to build a pipeline for the processing of structured data.
+
+Structured data is represented by the Record interface. LogLine implemets that interface and
```
arl, line 4:
```suggestion
Structured data is represented by the Record interface. LogLine implements that interface and
```

config.go
```diff
@@ -58,12 +62,13 @@ type ConfigFilter struct {
...
+	// Procs defines the number of baker outputs running concurrently.
+	// Only set Procs to a value greater than 1 if the output is concurrent safe.
 	Procs         int
-	ChanSize      int
-	RawChanSize   int
-	Sharding      string
-	Fields        []string
+	ChanSize      int      // ChanSize represents the size of the channel to send records to the ouput component(s), the default value is 16384
+	Sharding      string   // field name to be used for sharding
+	Fields        []string // fields to provide to the output as values
```
arl, line 71: Not making a true github suggestion since this would probably fail on ci because of gofmt
```
	Sharding      string   // Sharding is the name of the field used for sharding
	Fields        []string // Fields holds the name of the record fields the output receives
```

topology.go
```diff
@@ -11,11 +11,13 @@ import (
...
+// how to process them (filter), and where to output the results (output+upload)
 type Topology struct {
-	Input   Input
-	Filters []Filter
-	Output  []Output
-	Upload  Upload
+	Input   Input    // The input component
+	Filters []Filter // The list of filters
+	Output  []Output // The output component (it's a slice because can be sharded)
+	Upload  Upload   // The optional upload component
```
arl, line 20: This is a case where the convention would be to write something like:

Input is the input component
Filters are the filters components
etc.

In this case it's admitted to not document the fields and that's what I think we could do since the comments wouldn't add any information.

however the Output comment is not correct. It's not a slice because it can be sharded, it's a slice because you can set procs>1 so I think we could ommit the comment also here, as you prefer

arl (approved):
Last round for me :+1: 
you can merge this once this last set of comments has been adressed

PR #63, opened 2020-10-26
Logline specialization

arl:
#### :question: What

Make `baker.LogLine` more easily customizable.
LogLine hardcoded constants are now only used in `logline.go` so that if one copies that file to their project all continues to work.
Document how to do it in `README.md`

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ supports any single-byte field separator and doesn't handle quotes (neither
 single nor double). The maximum number of fields is hard-coded by the
 `LogLineNumFields` constant which is 3000. 100 extra fields can be stored at
 runtime in a `LogLine` (also hardcoded with `NumFieldsBaker`), these extra fields
-are a fast way to exchange between filters and/or outputs but they are neither
+are a fast way to exchange data between filters and/or outputs but they are neither
 handled during `Parsing` (i.e `LogLine.Parse`) nor serialization (`LogLine.ToText`).
 
 If the hardcoded values for `LogLineNumFields` and `NumFieldsBaker` do not suit
--- a/README.md
+++ b/README.md
@@ -34,7 +34,9 @@ Baker is fully parallel and maximizes usage of both CPU-bound and I/O bound pipe
   - [Provided Baker components](#provided-baker-components)
     - [Inputs](#inputs-1)
       - [KCL](#kcl)
-        - [Implementation and prevent throttling prevention](#implementation-and-prevent-throttling-prevention)
+        - [Implementation and throttling prevention](#implementation-and-throttling-prevention)
+  - [Working with baker.Record](#working-with-bakerrecord)
+    - [`baker.LogLine` CSV record](#bakerlogline-csv-record)
   - [Tuning parallelism](#tuning-parallelism)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -452,6 +454,39 @@ Doing so, we're guaranteed to never exceed the per-shard read througput limit of
...
+At the moment, `baker` proposes an unique `Record` implementation, `baker.LogLine`.
+
+### `baker.LogLine` CSV record
+
+`baker.LogLine` is an highly optimized CSV compatible Record implementation. It 
+supports any single-byte field separator and doesn't handle quotes (neither 
+single nor double). The maximum number of fields is hard-coded by the
+`LogLineNumFields` constant which is 3000. 100 extra fields can be stored at
+runtime in a `LogLine` (also hardcoded with `NumFieldsBaker`), these extra fields
+are a fast way to exchange between filters and/or outputs but they are neither
```
tommyblue, line 471:
```suggestion
are a fast way to exchange data between filters and/or outputs but they are neither
```

README.md
```diff
@@ -452,6 +454,39 @@ Doing so, we're guaranteed to never exceed the per-shard read througput limit of
...
+handled during `Parsing` (i.e `LogLine.Parse`) nor serialization (`LogLine.ToText`).
+
+If the hardcoded values for `LogLineNumFields` and `NumFieldsBaker` do not suit
+your needs, it's advised that you copy `logline.go` in your project and modify
+the constants declared at the top of the file. Your specialized `LogLine` will
```
tommyblue, line 476: the user should copy the logline but also rename the object, if I'm not mistaken. If true, I'd also change the code example below using something like `MyCustomLogline`

stats_test.go
```diff
@@ -110,3 +112,110 @@ func TestStatsDumper(t *testing.T) {
 
 	testutil.DiffWithGolden(t, buf.Bytes(), golden)
 }
+
+func TestStatsDumperInvalidRecords(t *testing.T) {
+	// This test contols the correct integration of the StatsDumper with the
```
tommyblue, line 117:
```suggestion
	// This test controls the correct integration of the StatsDumper with the
```

topology.go
```diff
@@ -369,8 +372,11 @@ func (t *Topology) runFilterChain() {
 			// Validate against patterns
 			if t.validate != nil {
 				// call external validation function
-				if ok, idx := t.validate(record); !ok {
-					atomic.AddInt64(&t.invalid[idx], 1)
+				ok, idx := t.validate(record)
+				if !ok {
```
tommyblue, line 376: is it as fast as the previous version?

tommyblue (approved):
lgtm

README.md
arl: No it's not necessary since the user will do this in its own project so in a different package so the definitions are never going to clash

topology.go
arl: No it's not since incrementing an atomic is faster than locking a mutex. However this is the only correct way to do it. Moreover the change is only seen when there's an invalid logline, the happy path stays the same

PR #64, opened 2020-10-27
logline.Copy didn't copy the idx

tommyblue:
#### :question: What

There was a bug in `logline.Copy`: when a logline were parsed and immediately copied (without setting any field), then the copy was wrong because `idx` was missing

I benchmarked the solution in this pr agains copying the whole logline (`*cpy = *l`) and this solution is a bit faster (~23000 ns/op vs ~26000 ns/op)

### DIFF
--- a/test_helper.go
+++ b/test_helper.go
@@ -97,6 +97,34 @@ func RecordConformanceTest(t *testing.T, create func() Record) {
 		}
 	})
 
+	t.Run("copy just parsed + Get", func(t *testing.T) {
+		org := create()
+		org.Set(0, []byte("foo"))
+		org.Set(1, []byte("bar"))
+		org.Set(260, []byte("baz"))
+
+		text := org.ToText(nil)
+
+		record := create()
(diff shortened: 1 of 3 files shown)


test_helper.go
```diff
@@ -97,6 +97,36 @@ func RecordConformanceTest(t *testing.T, create func() Record) {
...
+	t.Run("copy just parsed + Get", func(t *testing.T) {
+		values := map[FieldIndex]string{
+			0:   "foo",
+			1:   "bar",
+			260: "baz",
+		}
+		org := create()
+		for k, v := range values {
+			org.Set(k, []byte(v))
+		}
```
arl, line 109: Not sure this is more easy/readable to create a map just to loop on it, rather than calling 3 times `org.Set()`
and 3 time `bytes.Equal` at the end of the test.
Also for consistency with other subtests

arl (approved):
Good catch. LGTM
Just a readability/style comment about the test

PR #66, opened 2020-11-02
doc: tuning concurrency

tommyblue:
#### :question: What

How to tune concurrency in baker

### DIFF
--- a/content/en/docs/Core concepts/concurrency.md
+++ b/content/en/docs/Core concepts/concurrency.md
@@ -4,28 +4,32 @@ date: 2020-11-02
 weight: 3
 ---
 
-Baker filters and ouput can be configured with a high concurrency profile (input doesn't support
-parallel processing).
+Baker allows to tune concurrency at various levels of a pipeline:
 
-This means that those components will process Records in parallel during the pipeline,
-improving performances but adding a small cons: Records order isn't guaranteed anymore
-(due to possible different speed processing between concurrent components).
+* input: Baker configuration doesn't expose knobs to tune input concurrency as it highly depends
+on the input source and how the input is implemented
+* filters: Baker runs N concurrent filter chains
+* output: Baker runs M concurrent outputs
 
-{{% alert title="Default behaviour" color="primary" %}}
-The default behaviour of Baker is to use concurrency for both filters and output.
-{{% /alert %}}
+By default then, Baker processes records concurrently, without any guaranteed order.  
+However if you need to maintain the order of the records through the whole pipeline, it is still
+possible disabling concurrency ([see below for details](#guarantee-records-order)).
 
-### Concurrent filtering
+### Filter chain concurrency
 
-The filters concurrency can be set defining the `procs` key in the `[filterchain]` section:
+The filter chain is a synchronous list of filters and the order of filters in the chain depends
+on the topology configuration in the [TOML file](/docs/core-concepts/toml/).
+
+By default, though, Baker executes multiple concurrent filter chains (the default value is 16)
+
+Filterchain concurrency can be set defining the `procs` key in the `[filterchain]` section:
 
 ```toml
 [filterchain]
 procs=16
 ```
 
-The default value is **16**.  
-To disable concurrency, set **procs=1**.
+Setting the value to **procs=1** disables the filter chain concurrency.
 
 ### Concurrent output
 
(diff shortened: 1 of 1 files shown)


content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
+---
+title: "Tuning concurrency"
+date: 2020-11-02
+weight: 3
+---
+
+Baker filters and ouput can be configured with a high concurrency profile (input doesn't support
+parallel processing).
```
arl, line 8: Not sure what does a high concurrency profile means.
Also saying that input doesn't support parallel processing (I'd use concurrent rather than parallel) is a bit misleading since in inputs there's no record processing per se. Also, inputs can be concurrent if you program them to be, inputs can concurrently fetch record, that's what the KCL input does for example.

Rough suggestion:

Baker allows to tune concurrency at various levels of a pipeline:
 - inputs: baker configuration doesn't expose knobs that are input-specific with respect to concurrent as it highly depends on the input source and how the input is implemented
 - filter: baker runs N concurrent filter chains. N default is ? and can be tuned with `procs`
 - output: baker runs N concurrent output. N default is ? and can be tuned with  `procs`

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+
+This means that those components will process Records in parallel during the pipeline,
+improving performances but adding a small cons: Records order isn't guaranteed anymore
```
arl, line 11: use of future tense.

Also, this seems a bit marketing, but I'd  prefer if we say that by default records are concurrently processed, without any ordering guarantee. However **you can** guarantee ordering by making the whole pipeline synchronous (procs=1 everywhere)

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+(due to possible different speed processing between concurrent components).
+
+{{% alert title="Default behaviour" color="primary" %}}
+The default behaviour of Baker is to use concurrency for both filters and output.
+{{% /alert %}}
+
+### Concurrent filtering
```
arl, line 18:
```suggestion
### Filter chain concurrency
```

Emphasize on the fact that a filter chain is a **synchronous** list of filters. But we can add more filter chain and so the filter chains as a whole become concurrent.

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+
+The filters concurrency can be set defining the `procs` key in the `[filterchain]` section:
```
arl, line 20:
```suggestion
Filterchain concurrency can be set defining the `procs` key in the `[filterchain]` section:
```

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+[output]
+procs=32
+```
+
+The default value is **32**.  
+To disable concurrency, set **procs=1**.
+
+#### Output concurrency support
+
+At the moment it is not possible to know whether an output supports concurrency. A good guess is to
```
arl, line 44: Uhm not sure I like that we don't know that and we have to guess :-)
Outputs should definitely document if they support concurrency or not. Speaking about sharding here adds a bit of confusion IMHO.

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+see if the can support sharding (the `CanShard()` function), but the proper way is to know the
```
arl, line 45: typo or missing word at the beggining of the sentence

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,59 @@
...
+If an output doesn't support concurrency, you should use `procs=1` to avoid corrupted output or
+lost data.
+
+We'll soon add a new function to the output to declare its support for concurrency, and Baker will
+return an error if `procs>1` is used with an output that doesn't support it.
+
+### Guarantee Records order
+
+Although it's not the primary goal of Baker, it is still possible to disable concurrency and thus
+guarantee that the order of Records coming from the input is maintained in the output.
```
arl, line 57:
```suggestion
guarantee records ordering from input through output.
```

arl (approved):

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,64 @@
...
+Baker allows to tune concurrency at various levels of a pipeline:
+
+* input: Baker configuration doesn't expose knobs to tune input concurrency as it highly depends
+on the input source and how the input is implemented
+* filters: Baker runs N concurrent filter chains
+* output: Baker runs M concurrent outputs
+
+By default then, Baker processes records concurrently, without any guaranteed order.  
+However if you need to maintain the order of the records through the whole pipeline, it is still
+possible disabling concurrency ([see below for details](#guarantee-records-order)).
```
arl, line 16: maybe: 
possible to disable concurrency
or
possible by disabling concurrency
?

content/en/docs/Core concepts/concurrency.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Filter chain concurrency
+
+The filter chain is a synchronous list of filters and the order of filters in the chain depends
```
arl, line 20:
```suggestion
The filter chain is a synchronous list of filters which are applied in the order in which they are listed in the topology configuration
```
seems a bit lighter to read IMHO

arl (approved):
Ok, 2 minor propositions
but LGTM!

PR #67, opened 2020-11-03
Doc dockerfile

arl:
#### :question: What

Add Dockerfile.base and Dockerfile, some makefile targets to use them and update the README for development and running the website server.

#### :hammer: How to test

1. `make docker-dev`
2. point your browser to http://locahost:1313/baker
3. modify `content/en/docs/Core concepts/toml.md`
4. the browser page should update

or

1. `make docker-run-prod`
2. point your browser to http://locahost/baker # port 80

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,16 +15,10 @@ build: ## Build the static files of the website
 	hugo
 
 docker-base: 
-	docker build -f Dockerfile.base -t baker-docs:base .
+	docker build -t baker-docs:base .
 
 docker-dev: docker-base ## Use docker for baker website development
-	docker run -w /baker -v $$PWD:/baker -p 1313:1313 baker-docs
-
-docker-build-prod: docker-base
-	docker build -t baker-docs:prod .
-
-docker-run-prod: docker-build-prod ## Build baker website production version and runs hugo server on port 80
-	docker run -p 80:1313 -t baker-docs:prod
+	docker run -w /baker -v $$PWD:/baker -p 1313:1313 -it baker-docs:base hugo server --bind=0.0.0.0
 
 .PHONY: docker-base
 
--- a/Makefile
+++ b/Makefile
@@ -13,3 +13,18 @@ dev: ## Run local server to check you content while writing
 
 build: ## Build the static files of the website
 	hugo
+
+docker-base: 
+	docker build -f Dockerfile.base -t baker-docs:base .
+
+docker-dev: docker-base ## Use docker for baker website development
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -13,3 +13,18 @@ dev: ## Run local server to check you content while writing
 
 build: ## Build the static files of the website
 	hugo
+
+docker-base: 
+	docker build -f Dockerfile.base -t baker-docs:base .
+
+docker-dev: docker-base ## Use docker for baker website development
+	docker run -w /baker -v $$PWD:/baker -p 1313:1313 baker-docs
```
tommyblue, line 21:
```suggestion
	docker run -w /baker -v $$PWD:/baker -p 1313:1313 baker-docs:base hugo server --bind=0.0.0.0
```

tommyblue: it doesn't work without this change

Makefile
```diff
@@ -13,3 +13,18 @@ dev: ## Run local server to check you content while writing
...
+
+docker-build-prod: docker-base
+	docker build -t baker-docs:prod .
+
+docker-run-prod: docker-build-prod ## Build baker website production version and runs hugo server on port 80
+	docker run -p 80:1313 -t baker-docs:prod
```
tommyblue, line 27: can we use a port different from 80? it fails on my machine as it is used

tommyblue: moreover, I'm not sure I'm getting the use of the prod version. It is only on the port 80 instead of 1313 with minified files. Maybe we should add a `build` target to build the website with docker?

tommyblue (changes requested):
please fix errors

Makefile
arl: I just forgot to add the entry-point in Dockerfile.base. done

Makefile
arl: > can we use a port different from 80? it fails on my machine as it is used

I thought we wanted a production version of the website, bound to port 80, to be deployed on aws inside a docker image.
If that's not what we want, then we can just remove `docker-build-prod` and `docker-run-prod` makefile targets altogether.

> Maybe we should add a build target to build the website with docker?

If we only want to use docker to build the website into a target directory, it's probably enough to bind mount the current directory (as does `docker-dev`).

tommyblue: yup. The production version is created with `make build` that creates static files to be uploaded to GH pages

arl: Ok so in this case I'm going to remove code and files related to `docker-prod` targets and add `docker-build` instead.
It will just use docker to build the website locally

tommyblue (approved):

PR #70, opened 2020-11-03
add intro and baker.Components pages

tommyblue:
#### :question: What

* Small intro to the documentation
* `baker.Components` doc page

### DIFF
--- a/content/en/docs/Core concepts/sharding.md
+++ b/content/en/docs/Core concepts/sharding.md
@@ -1,0 +1,23 @@
+---
+title: "Sharding"
+date: 2020-11-02
+weight: 500
+---
+
+# TBD - these are just notes
+
+The `[output.sharding]` configuration in the TOML file tells which field of the records must be
+used to calculate the sharding.
+
+The field name, transformed to `FieldIndex` thanks to the `FieldByName` function (see
+[below](#fieldbyname)), is the index of the map that corresponds to a sharding function.
+
+The sharding function receives a Record and returns a sharding value (`uint64`) that Baker
+uses to send the Record to an output process.
+
+{{% alert color="info" %}}
+How the sharding value is calculated is up to the function, but it should try to have a linear
+distribution of records (to evenly distribute the load among the output concurrent processes)
+and, although not required, it should also use value of the field that has been configured in
+`[output.sharding]`.
+{{% /alert %}}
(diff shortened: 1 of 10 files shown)


content/en/docs/Core concepts/baker_components.md
```diff
@@ -0,0 +1,135 @@
...
+
+To learn more about this topic, read the
+[dedicated section](/docs/core-concepts/toml/#user-defined-configurations) in the Pipeline
+configuration page.
+
+## ShardingFuncs
+
+This field is a dictionary of functions used to calculate how to shard records among the
+available output goroutines (see [Tuning concurrency](/docs/core-concepts/concurrency) for
+details about output concurrency).
```
arl: The sentence is not super clear IMHO, what do you think of something along these lines?
```suggestion
This field holds a dictionary associating field indices to hash functions. When sharding
is enabled, these hash functions are used to determine which shard a record is sent to.
```

Also, I'm not sure we should then go into such details about sharding here since we're in the `baker.Components` section, IMHO the rest of the ShardingFunc section should reside in the sharding guide and/or sharding core concept pages, we should keep it very close to what  the baker.Components field does.

content/en/docs/Core concepts/baker_components.md
```diff
@@ -0,0 +1,135 @@
...
+
+The `[output.sharding]` configuration in the TOML file tells which field of the records must be
+used to calculate the sharding.
```
arl: These details should probably be in the sharding concept/use page

content/en/docs/Core concepts/baker_components.md
```diff
@@ -0,0 +1,135 @@
...
+	Validate       ValidationFunc
+	CreateRecord   func() Record
+	FieldByName    func(string) (FieldIndex, bool)
+	FieldName      func(FieldIndex) string
+}
+```
+
+## Inputs, Filters, Outputs and Uploads
+
+These fields contain the list of components that are available to the topology.
```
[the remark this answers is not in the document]
tommyblue: I like the idea of a more structured page where we can add more details, like in this case, instead of a list with short comments. Please check now after the new commits

content/en/docs/Core concepts/sharding.md
```diff
@@ -0,0 +1,23 @@
+---
```
tommyblue, line 1: I put here the sentences I extracted from the other file after comments, just to avoid losing them

arl (approved):
lgtm

PR #73, opened 2020-11-05
Create a MainCLI fn with utilities

tommyblue:
#### :question: What

Add the `MainCLI` function to help building a Baker main with some already included features


baker_cli.go
```diff
@@ -0,0 +1,181 @@
...
+// MainCLI provides a handy way to quickly create a Baker program only providing
+// the available Components.
+//
+// The function includes many utilities that can be configured by command line arguments:
+// -help: Prints available options and components
+// -version: print build version (build with `-ldflags="-X 'github.com/AdRoll/baker.BuildVersion=someversion'"` to set the value)
+// -v: verbose logging (not compatible with -q)
+// -q: quiet logging (not compatible with -v)
+// -pretty: logs in textual format instead of JSON format
+// -pprof: run a pprof server on the provided host:port address
```
arl, line 38: In generated go doc, `// hello` and `//  hello` (2 spaces) are formatted differently.
With 2 spaces between `//` and the text, the text shows as code block in godoc. Using a code block for the list of cli arguments here could maybe look better

baker_cli.go
```diff
@@ -0,0 +1,181 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	flagHelpConfig = flag.String("help", "", "show help for a `component` (input/filter/output/upload) (use '*' to dump all)")
+	flagVersion    = flag.Bool("version", false, "print build version number")
+	flagVerbose    = flag.Bool("v", false, "verbose logging (debug level)")
+	flagQuiet      = flag.Bool("q", false, "quiet logging (warn level)")
+	flagPretty     = flag.Bool("pretty", false, "human-readable logging (unstructured logging)")
+	flagPProf      = flag.String("pprof", "", `run pprof server on host port provided (disabled if ""), use "localhost:"  for a free port`)
```
arl, line 23: can we declare flags in mainCLI instead of using global variables?

baker_cli.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+// Use `-ldflags="-X 'github.com/AdRoll/baker.BuildVersion=someversion'"` when building Baker to set this value
+var BuildVersion = "-- unknown --"
+
+// MainCLI provides a handy way to quickly create a Baker program only providing
+// the available Components.
```
arl, line 30:
```suggestion
// MainCLI starts provides a handy way to quickly create a command-line interface to Baker 
// by providing the list of components available to build and run a topology.
```

baker_cli.go
```diff
@@ -0,0 +1,181 @@
...
+//
+// The function includes many utilities that can be configured by command line arguments:
+//  -help: Prints available options and components
+//  -version: print build version (build with `-ldflags="-X 'github.com/AdRoll/baker.BuildVersion=someversion'"` to set the value)
```
arl, line 34: not sure we should include the version flag here. this is a more advanced use, requiring to build baker with specific flag.
Note that flags could also been declared by whatever main application calling MainCLI

baker_cli.go
arl: I'd add somewhere that `./baker topology.toml` runs the topology.

baker_cli.go
```diff
@@ -0,0 +1,181 @@
...
+	f, err := os.Open(flag.Arg(0))
+	if err != nil {
+		return fmt.Errorf("errors opening config: %v", err)
+	}
+
+	cfg, err := NewConfigFromToml(f, components)
+	if err != nil {
+		return err
+	}
+	f.Close()
```
arl, line 100: this should be `defer` and done earlier, just after a successful `os.Open`

examples/advanced/main.go
```diff
@@ -1,47 +1,86 @@
-// advanced example shows a complex program using almost all baker features
+// advanced example shows a complex setup of baker.Components
```
arl, line 1: IMHO _advanced_ seems better than _complex_ here

examples/advanced/main.go
```diff
@@ -1,47 +1,86 @@
...
+	}
+}
 
-	flag.Usage = displayProgramUsage
-	flag.Parse()
+func simpleHash(r baker.Record, idx baker.FieldIndex) uint64 {
+	f := fnv.New64()
+	f.Write(r.Get(idx))
+	return f.Sum64()
+}
```
arl, line 25: maybe we should move simpleHash close to where it's used (bottom)

baker_cli.go
tommyblue: how would you show the version number? if the `main()` declares a flag, it can't use it until `flag.Parse()` is called in `mainCLI`, but once called then the control is in the hand of `mainCLI` and the original `main` fn can't add options for the flags it declared (like printing the version and exit).

tommyblue: maybe one idea, though a bit cumbersome (or cucumbersome :laughing: ) could be to define some functions in the `main` and pass them to `MainCLI`, that calls them after parse:

```go
// main()
var fns = []func(){
	func() {
		if *flagVersion {
			fmt.Printf("Baker version: %s\n", build)
			os.Exit(0)
		}
	},
}
baker.MainCLI(components, fns)

// MainCLI
func MainCLI(components Components, closures []func()) error {
	flag.Parse()
	for _, fn := range closures {
		fn()
	}
}
```

arl: I haven't thought about that when I wrote my first comment.
I'm not crazy about passing additional functions. 
I think MainCLi should be enough for most cases, if one needs more, they can just copy baker.MainCLI into their code and modify it to suit their needs, seems reasonable enough

examples/advanced/main.go
arl: you don't agree? @tommyblue

arl (approved):
LGTM :+1: 

There's one of my previous _nitpicking_ comments that hasn't been addressed, maybe on purpose.

baker_cli.go
arl: @tommyblue also this hasn't been addressed yet
The first non positional CLI argument is going to be interpreted as the path to the topology file.

arl:
@tommyblue There's another comment that hasn't been addressed
 - comment about `baker topology.toml`
 - s/complex/advanced

tommyblue:
uhm, somethings going on there, I addressed that. let me check again

tommyblue:
ok,done

PR #76, opened 2020-11-06
doc: How to create a filter

tommyblue:
#### :question: What

How to create a filter

### DIFF
--- a/content/en/docs/How To/create_filter.md
+++ b/content/en/docs/How To/create_filter.md
@@ -1,0 +1,208 @@
+---
+title: "Create a custom filter component"
+date: 2020-11-02
+weight: 700
+---
+
+Creating a custom filter is probably the most common action a Baker user will perform.
+
+In fact, filters are the components that apply the business logic to a Baker pipeline,
+**creating or discarding records or modifying fields.**
+
+A working example of a custom filter can be found in the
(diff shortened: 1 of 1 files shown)


content/en/docs/How To/create_filter.md
```diff
@@ -0,0 +1,208 @@
...
+    // Call next the 1st time
+    next(r)
+
+    // WRONG, it is the same record as above
+    next(r)
+
+    // CORRECT, this is a copy of the record
+    next(r.Copy())
+}
+```
```
arl, line 182: It's a good idea to illustrate the case of duplicated records but I'd be more explicit about what we're illustrating here as it's not super obvious. Maybe add something like:

> Filters can duplicate incoming records, and thus have more records come out than records that came in.

content/en/docs/How To/create_filter.md
```diff
@@ -0,0 +1,208 @@
...
+
+### CreateRecord()
+
+A new, empty, record is created calling the `CreateRecord` function.
+The `CreateRecord` function is available as part of the
+[FilterParams](https://pkg.go.dev/github.com/AdRoll/baker#FilterParams) argument of the
+[filter constructor](#filter-constructor). If you plan to use it in the `Process` function
+than store it to the filter object in the constructor as shown in this example:
```
arl, line 190: s/than/then

arl (approved):

PR #77, opened 2020-11-10
doc: how to create an input

tommyblue:
#### :question: What

Doc: how-to create an input component

### DIFF
--- a/content/en/docs/How To/create_input.md
+++ b/content/en/docs/How To/create_input.md
@@ -33,13 +33,9 @@ The [`Data`](https://pkg.go.dev/github.com/AdRoll/baker#Data) object that the in
 with read data has two fields: `Bytes` that must contain the raw read bytes (possible contains
 more records) and `Meta`.
 
-[`Metadata`](https://pkg.go.dev/github.com/AdRoll/baker#Metadata) can contain any information
-that the input wants to share with the other components of the pipeline about the chunk of bytes
-produced.  
-Typical information could be the timestamp of the read, details about the source, etc. 
-
-Those metadata will be associated to each record extracted from `Bytes` and will be available
-to the whole pipeline.
+[`Metadata`](https://pkg.go.dev/github.com/AdRoll/baker#Metadata) can contain additional 
+information Baker will associate with each of the serialized Record contained in `Data`.  
+Typical information could be the time of retrieval, the filename (in case `Records` come from a file), etc.
 
 ## The Input interface
 
--- a/content/en/docs/How To/create_input.md
+++ b/content/en/docs/How To/create_input.md
@@ -17,7 +17,6 @@ To create an input and make it available to Baker, one must:
 * Implement the [Input](https://pkg.go.dev/github.com/AdRoll/baker#Input) interface
 * Fill an [`InputDesc`](https://pkg.go.dev/github.com/AdRoll/baker#InputDesc) structure and register it
 within Baker via [`Components`](https://pkg.go.dev/github.com/AdRoll/baker#Components).
-the available inputs in [`Components`](https://pkg.go.dev/github.com/AdRoll/baker#Components)
 
 ## Daemon vs Batch
 
--- a/content/en/docs/How To/create_input.md
+++ b/content/en/docs/How To/create_input.md
@@ -22,7 +22,7 @@ the available inputs in [`Components`](https://pkg.go.dev/github.com/AdRoll/bake
 
 The input component determines the Baker behavior between a batch processor or a long-living daemon.
 
-If the input exits when its data processing has completed, then Baker waits the topology to end
+If the input exits when its data processing has completed, then Baker waits for the topology to end
 and then exits.
 
 If the input never exits, then Baker acts as a daemon.
--- a/content/en/docs/How To/create_input.md
+++ b/content/en/docs/How To/create_input.md
@@ -1,0 +1,102 @@
+---
+title: "Create a custom input component"
+date: 2020-11-02
+weight: 650
+---
+The input component goal is to retrieve Records in form of bytes and send them to the filterchain.
+
+The input isn't in charge of splitting/parsing the input data into Records (that is done by Baker),
+but only retrieving them as fast as possible in raw format adding, if any, metadata to them.
+
+The input shares a [`*Data`](https://pkg.go.dev/github.com/AdRoll/baker#Data) channel with the
+filter chain. The channel size is customizable with `[input] chansize=<value>` (default to 1024).
+
+To create an input and make it available to Baker, one must:
+
+* Implement the [Input](https://pkg.go.dev/github.com/AdRoll/baker#Input) interface
+* Add an [`InputDesc`](https://pkg.go.dev/github.com/AdRoll/baker#InputDesc) for the input to
+the available inputs in [`Components`](https://pkg.go.dev/github.com/AdRoll/baker#Components)
+
+## Daemon vs Batch
+
+The input component determines the Baker behavior between a batch processor or a long-living daemon.
+
+If the input exits when its data processing has completed, then Baker waits the topology to end
+and then exits.
+
+If the input never exits, then Baker acts as a daemon.
+
+## Data
+
+The [`Data`](https://pkg.go.dev/github.com/AdRoll/baker#Data) object that the input must fill in
+with read data has two fields: `Bytes` that must contain the raw read bytes (possible contains
+more records) and `Meta`.
+
+[`Metadata`](https://pkg.go.dev/github.com/AdRoll/baker#Metadata) can contain any information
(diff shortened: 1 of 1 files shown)


content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+If the input exits when its data processing has completed, then Baker waits the topology to end
+and then exits.
+
+If the input never exits, then Baker acts as a daemon.
+
+## Data
+
+The [`Data`](https://pkg.go.dev/github.com/AdRoll/baker#Data) object that the input must fill in
+with read data has two fields: `Bytes` that must contain the raw read bytes (possible contains
+more records) and `Meta`.
```
arl, line 33: Even if that may change in the future, we should say somewhere that `baker.Data.Bytes` must represent `\n`-separated un-serialized records

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+
+[`Metadata`](https://pkg.go.dev/github.com/AdRoll/baker#Metadata) can contain any information
+that the input wants to share with the other components of the pipeline about the chunk of bytes
+produced.  
+Typical information could be the timestamp of the read, details about the source, etc. 
+
+Those metadata will be associated to each record extracted from `Bytes` and will be available
+to the whole pipeline.
```
arl, line 41:
```suggestion
[`Metadata`](https://pkg.go.dev/github.com/AdRoll/baker#Metadata) can contain additional 
information Baker will associate with each of the serialized Record contains in `Data`.  
Typical information could be the time of retrieval, the filename (in case `Records` come a file), etc.
```

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+
+The `Run` function implements the component logic and receives a channel where it sends the
+[raw data](https://pkg.go.dev/github.com/AdRoll/baker#Data) it processes.
+
+`FreeMem` is the function called by Baker when a chunk of data has been completely processed and
+its memory can be used again. It's up to the input to choose whether to recycle it or not.  
+An often used pattern is to manage the objects with [`sync.Pool`](https://golang.org/pkg/sync/#Pool),
+see the [TCP](https://github.com/AdRoll/baker/blob/main/input/tcp.go),
+[List](https://github.com/AdRoll/baker/blob/main/input/list.go) or
+[SQS](https://github.com/AdRoll/baker/blob/main/input/sqs.go) inputs for inspiration.
```
arl, line 65:
```suggestion
`FreeMem(data *Data)` is called by Baker when `data` is no longer needed. This is an occasion
for the input to recycle memory, for example if the input uses a `sync.Pool` to create new 
instances of `baker.Data`. 
```

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+var MyInputDesc = baker.InputDesc{
+	Name:   "MyInput",
+	New:    NewMyInput,
+	Config: &MyInputConfig{},
+	Help:   "Help message for the input",
+}
+```
+
+This object has a `Name`, that is used in the Baker configuration file to identify the input,
+a costructor function (`New`), a config object (used to parse the input configuration in the
```
arl, line 79: `costructor` -> `constructor-like`

Why not use list-styled items here? one per field?

nit: the config object is not really _used to parse_ the input configuration, since this could be understood as if it was the Config which parses the TOML.

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+[SQS](https://github.com/AdRoll/baker/blob/main/input/sqs.go) inputs for inspiration.
+
+## InputDesc
+
+```go
+var MyInputDesc = baker.InputDesc{
+	Name:   "MyInput",
+	New:    NewMyInput,
+	Config: &MyInputConfig{},
+	Help:   "Help message for the input",
```
arl, line 74: Maybe instead of `Help message for the input`, something like: `High-level description of MyInput`?

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+}
+```
+
+This object has a `Name`, that is used in the Baker configuration file to identify the input,
+a costructor function (`New`), a config object (used to parse the input configuration in the
+TOML file) and a help text that must help the users to use the component and its configuration
+parameters.
+
+### Input constructor
```
arl, line 83: In Go these are conventionally called _constructor-like functions_, since they're not technically (nit) constructors, they're just free functions that happen to be used to construct a specific instances of a specific type. 

```suggestion
### Input constructor-like function
```

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+
+The `New` key in the `InputDesc` object represents the constructor function.
+
+The function receives a [InputParams](https://pkg.go.dev/github.com/AdRoll/baker#InputParams)
+object and returns an instance of [Input](https://pkg.go.dev/github.com/AdRoll/baker#Input).
+
+The function should verify the configuration params into `InputParams.DecodedConfig` and initialize
+the component. Additional operations can be performed in the `Run` function when called.
```
arl, line 91: I'm not sure of the added value of the last sentence, but maybe I didn't quite get what you meant.

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+### The input configuration and help
+
+The input configuration object (`MyInputConfig` in the previous example) must export all
+configuration parameters that the user can set in the TOML topology file.
+
+Each key in the struct must include a `help` string tag and a `required` boolean tag, both are
+mandatory.
+
+The former helps the user to understand the possible values of the field, the latter tells Baker
+whether to refuse a missing configuration param.
```
arl, line 102: :bulb:  **idea** :bulb: 
We should probably list somewhere the types we support for the fields of the Config structures, since we don't support every type there is.

Rather than describing those in each component page, we could have a page dedicated to the Config structs, in which we also list the supported struct tags (`required`, `help` and `default`).

And list this new page here and there.

What do you think?

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,102 @@
...
+title: "Create a custom input component"
+date: 2020-11-02
+weight: 650
+---
+The input component goal is to retrieve Records in form of bytes and send them to the filterchain.
+
+The input isn't in charge of splitting/parsing the input data into Records (that is done by Baker),
+but only retrieving them as fast as possible in raw format adding, if any, metadata to them.
+
+The input shares a [`*Data`](https://pkg.go.dev/github.com/AdRoll/baker#Data) channel with the
```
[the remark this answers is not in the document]
tommyblue: fixed and rephrased

content/en/docs/How To/create_input.md
tommyblue: added

content/en/docs/How To/create_input.md
tommyblue: I didn't use a list here to avoid replicating the code docs and to maintain a more colloquial style, as it's just a brief sentence without too many details

content/en/docs/How To/create_input.md
tommyblue: :+1:

content/en/docs/How To/create_input.md
tommyblue: well, I wanted to say that if something isn't feasible in the constructor-like function, one can also use the Run function, but yeah, it's not very clear so I'm removing it

content/en/docs/How To/create_input.md
tommyblue: sounds good, we can also simplify all these components page. I'll leave the page as it is now, when we'll create the suggested page we can rework the components pages removing duplicated info

content/en/docs/How To/create_input.md
```diff
@@ -0,0 +1,96 @@
...
+The function should verify the configuration params into `InputParams.DecodedConfig` and initialize
+the component.
+
+### The input configuration and help
+
+The input configuration object (`MyInputConfig` in the previous example) must export all
+configuration parameters that the user can set in the TOML topology file.
+
+Each key in the struct must include a `help` string tag and a `required` boolean tag, both are
+mandatory.
```
arl, line 93: `required` is not mandatory. I made this comment on https://github.com/AdRoll/baker/pull/78/files#r521546397 but I haven't seen it here the first time.
There's the same also on the upload PR

tommyblue: fixed!

tommyblue:
all done @arl

arl (approved):
:+1: LGTM

PR #78, opened 2020-11-10
doc: how to create an ouput

tommyblue:
#### :question: What

doc: how to create an output component

PR #79, opened 2020-11-11
doc: how to create an upload

tommyblue:
#### :question: What

doc: how to create an upload component

### DIFF
--- a/content/en/docs/How To/create_upload.md
+++ b/content/en/docs/How To/create_upload.md
@@ -49,7 +49,7 @@ configuration parameters.
 
 ### Upload constructor-like function
 
-The `New` key in the `UploadDesc` object represents the constructor-like function.
+The `New` key in the `UploadDesc` object should be set to a function returning an Upload.
 
 The function receives a [UploadParams](https://pkg.go.dev/github.com/AdRoll/baker#UploadParams)
 object and returns an instance of [Upload](https://pkg.go.dev/github.com/AdRoll/baker#Upload).
--- a/content/en/docs/How To/create_upload.md
+++ b/content/en/docs/How To/create_upload.md
@@ -71,7 +71,7 @@ is not defined.
 
 ## The files to upload
 
-Through the channel, the upload receives from the output paths to local files that it must uploads.
+Through the channel, the upload receives from the output paths to local files that it must upload.
 
 The only upload existing at the moment, S3, removes those files once uploaded, but there isn't a
 golden rule for what to do with them. This is up to the upload component and should be chosen
--- a/content/en/docs/How To/create_upload.md
+++ b/content/en/docs/How To/create_upload.md
@@ -1,0 +1,85 @@
+---
+title: "Create a custom upload component"
+date: 2020-11-11
+weight: 750
+---
+The last (optional) component of a Baker pipeline is the Upload, whose job is to, precisely,
+upload somewhere what the output component produces.
+
+To create an upload component and make it available to Baker, one must:
+
+* Implement the [Upload](https://pkg.go.dev/github.com/AdRoll/baker#Upload) interface
+* Fill an [`UploadDesc`](https://pkg.go.dev/github.com/AdRoll/baker#UploadDesc) structure and
+register it within Baker via [`Components`](https://pkg.go.dev/github.com/AdRoll/baker#Components).
+
+## The Upload interface
+
+```go
+type Upload interface {
+	Run(upch <-chan string) error
+	Stop()
+	Stats() UploadStats
+}
+```
+
+The [Upload interface](https://pkg.go.dev/github.com/AdRoll/baker#Upload) must be implemented when
+creating a new Upload component.
+
+The `Run` function implements the component logic and receives a channel where the output sends what
+it produces (most probably file paths).
+
+Currently, only the [S3](https://github.com/AdRoll/baker/blob/main/upload/s3.go) component exists
+and it expects to receive in the channel the path to the files produced by an output.
+
+## UploadDesc
+
+```go
+var MyUploadDesc = baker.UploadDesc{
+	Name:   "MyUpload",
+	New:    NewMyUpload,
+	Config: &MyUploadConfig{},
+	Help:   "High-level description of MyUpload",
+}
+```
+
+This object has a `Name`, that is used in the Baker configuration file to identify the upload,
+a costructor-like function (`New`), a config object (where the parsed upload configuration from the
+TOML file is stored) and a help text that must help the users to use the component and its
+configuration parameters.
+
+### Upload constructor-like function
+
+The `New` key in the `UploadDesc` object represents the constructor-like function.
+
+The function receives a [UploadParams](https://pkg.go.dev/github.com/AdRoll/baker#UploadParams)
+object and returns an instance of [Upload](https://pkg.go.dev/github.com/AdRoll/baker#Upload).
+
+The function should verify the configuration params into `UploadParams.DecodedConfig` and initialize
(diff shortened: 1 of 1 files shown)


content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+var MyUploadDesc = baker.UploadDesc{
+	Name:   "MyUpload",
+	New:    NewMyUpload,
+	Config: &MyUploadConfig{},
+	Help:   "High-level description of MyUpload",
+}
+```
+
+This object has a `Name`, that is used in the Baker configuration file to identify the upload,
+a costructor-like function (`New`), a config object (where the parsed upload configuration from the
```
arl, line 46:
```suggestion
a constructor-like function (`New`), a config object (where the parsed upload configuration from the
```

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+TOML file is stored) and a help text that must help the users to use the component and its
+configuration parameters.
+
+### Upload constructor-like function
+
+The `New` key in the `UploadDesc` object represents the constructor-like function.
```
arl, line 52: I feel that constructor-like it a bit too much repeated:
```suggestion
The `New` key in the `UploadDesc` object should be set to a function returning an Upload.
```

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+The function receives a [UploadParams](https://pkg.go.dev/github.com/AdRoll/baker#UploadParams)
```
arl, line 54: a -> an maybe?

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+object and returns an instance of [Upload](https://pkg.go.dev/github.com/AdRoll/baker#Upload).
+
+The function should verify the configuration params into `UploadParams.DecodedConfig` and initialize
```
arl, line 57: The function -> It:
```suggestion
It should verify the configuration, accessed via `UploadParams.DecodedConfig` and initialize
```

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+the component accordingly.
+
+### The upload configuration and help
```
arl, line 60:
```suggestion
### Upload configuration and help
```

?

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+The upload configuration object (`MyUploadConfig` in the previous example) must export all
+configuration parameters that the user can set in the TOML topology file.
+
+Each key in the struct must include a `help` string tag and a `required` boolean tag, both are
+mandatory.
```
arl, line 66: `required` is not mandatory. See https://github.com/AdRoll/baker/pull/78/files#r521546397
Same in all PRs actually

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+The former helps the user to understand the possible values of the field, the latter tells Baker
+whether to refuse a missing configuration param.
```
arl, line 69: There's a suggestion in https://github.com/AdRoll/baker/pull/78/files#r521546397
about that

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+## The shared channel
```
arl, line 71: Not sure we should say use the word shared here for the channel. A channel is by definition shared (or it's useless), so it find it slightly confusing.

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Create a custom upload component"
+date: 2020-11-11
+weight: 750
+---
+The last (optional) component of a Baker pipeline is the Upload, whose job is to, precisely,
+upload somewhere what the output component produces.
```
arl, line 7: I think it should be clear from the start that the upload **must** deal with absolute paths to the local filesystem. There's no way around that, since that's how outputs that produces files work, and that's how the S3 upload works.
If I was to write an output which doesn't write absolute paths to the upload channel, then we'd have a problem since this output couldn't be placed before a S3 upload, which would defeat the modular purpose of baker component.

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+Each key in the struct must include a `help` string tag and a `required` boolean tag, both are
+mandatory.
+
+The former helps the user to understand the possible values of the field, the latter tells Baker
+whether to refuse a missing configuration param.
+
+## The shared channel
+
+What the upload component receives in the string channel is, indeed, a string and, at the moment,
+the only upload implementation (S3) expects to find there the path to local files.
```
arl, line 74: As I've written (in the top comment of this review) upload components contract says they must handle absolute path to the local filesystem, so saying the only upload implementation expects to find that seems to mean that there could be another way for which an upload can expect something else, which there couldn't.

Also, if this a string channel, it's indeed a string it receives ;-) so this is a bit stating the obvious

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+The S3 upload removes those files once uploaded, but there isn't a golden rule for what to do with
+them. This is up to the upload component and should be chosen wisely and documented extensively.
```
arl, line 77: :+1:

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,85 @@
...
+
+At the same time, it is not an obligation for the string sent into the channel to represent
+a local file path and thus it is also possible that an upload component is not compatible with an
+output component (depending on what is sent to the channel and what is expected).
+
+Again, as there is not a golden rule about what is sent to the channel, the best thing to do for
+both the output and upload components is to create good documentation to inform the users of their
+behaviors.
```
arl, line 85: :-1: Not sure we should advertise that. It's true that tomorrow I could create an output that doesn't send paths through `upch` but, say, id's, and connect it to an upload that know how to deal with this. But, I honestly can't find a good reason for which the upload behavior in this case couldn't be performed directly in the output

tommyblue: agreed

tommyblue (commented):
all done

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,78 @@
...
+Each field in the struct must include a `help` string tag (mandatory) and a `required` boolean tag
+(default to `false`).
+
+All these parameters appear in the generated help. `help` should describe the parameter role and/or
+its possible values, `required` informs Baker it should refuse configurations in which that field
+is not defined.
+
+## The files to upload
+
+Through the channel, the upload receives from the output paths to local files that it must uploads.
```
arl, line 74:
```suggestion
Through the channel, the upload receives from the output paths to local files that it must upload.
```

content/en/docs/How To/create_upload.md
```diff
@@ -0,0 +1,78 @@
...
+
+The only upload existing at the moment, S3, removes those files once uploaded, but there isn't a
```
arl, line 76:
```suggestion
The only Upload component implemented at the moment, S3, removes those files once uploaded, but there isn't a
```

arl (approved):
I just let 2 minor comments but it looks good to me
:-1:

PR #84, opened 2020-11-13
Read field mapping from toml

arl:
#### :question: What

This is another step in the direction of making baker a full-featured binary one can use without touching code.
This gives the possibilty to defined field mapping in TOML. What we call fIeld mapping is a two-way mapping from field indexes and field names.  
Before this change, one had to provide 2 function `FieldName` and `FieldByName` into baker.Components, even for super basic use of baker. Now you can specify this in TOML, from which 2 closures are generated and assigned into these 2 functions. This is done in `NewConfigFromTOML`.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/config.go
+++ b/config.go
@@ -399,12 +399,14 @@ func NewConfigFromToml(f io.Reader, comp Components) (*Config, error) {
 		return nil, fmt.Errorf("invalid keys in configuration file: %v", keys)
 	}
 
+	if err := assignFieldMapping(&cfg, comp); err != nil {
+		return nil, err
+	}
+
 	// Copy pluggable functions
 	cfg.shardingFuncs = comp.ShardingFuncs
 	cfg.validate = comp.Validate
 	cfg.createRecord = comp.CreateRecord
-	cfg.fieldByName = comp.FieldByName
-	cfg.fieldName = comp.FieldName
 
 	// Fill-in with missing defaults
 	return &cfg, cfg.fillDefaults()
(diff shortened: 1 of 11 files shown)


config.go
```diff
@@ -406,6 +418,51 @@ func hasConfig(cfg interface{}) bool {
...
+	// First, get inconsistent cases out of the way.
+	if len(cfg.Fields.Names) == 0 && comp.FieldByName == nil && comp.FieldName == nil {
+		return fmt.Errorf("field indexes/names have not been set")
+	}
+
+	if (comp.FieldByName == nil) != (comp.FieldName == nil) {
+		return fmt.Errorf("FieldByName and FieldName must be either both set or both unset")
+	}
+
+	if len(cfg.Fields.Names) != 0 && comp.FieldByName != nil && comp.FieldName != nil {
```
tommyblue, line 434: shouldn't this be `a && (B || C)`?

examples/basic/main.go
```diff
@@ -45,9 +54,10 @@ procs=1
 	PathString="./_out/output.csv.gz"
 `
 	c := baker.Components{
-		Inputs:      input.All,
-		Outputs:     output.All,
-		FieldByName: fieldByName,
+		Inputs:  input.All,
+		Outputs: output.All,
+		Filters: filter.All,
+		// FieldByName: fieldByName,
```
tommyblue, line 60: leftover

config.go
arl: At this point we already know that `B == C` (they're either both true or both false, cf L430).

examples/basic/main.go
arl: done

config.go
tommyblue: ok, agree. But still I'm not sure why we have `comp.FieldByName != nil && comp.FieldName != nil` here.

I mean: or we check `a && (B || C)` (because `a && B && C` is incorrect) or we can remove the `&& comp.FieldName != nil` part, no?

I say **incorrect** because the only chance to hit the 3rd part of the `if`, (i.e. `&& comp.FieldName != nil`) is if the previous check (L430) is removed. But, in that case, `A && B && C` is a wrong check (`a==4, b==nil, c==&someptr` doesn't trigger the error)

tommyblue (approved):
resolved the `if`, lgtm

config.go
arl: I thought about that, the last check is surely not needed, but I wanted to keep the symmetry between both functions checks because it seemed misleading to me to only check for one of them.
I think I found a solution to make the code more easily readable

arl: done

tommyblue:
perfect

PR #85, opened 2020-11-16
doc: basic tutorial

tommyblue:
#### :question: What

Create a basic tutorial. Also adds a small `how to read stats` page as it is linked in the tutorial

PR #86, opened 2020-11-16
Basic tutorial files

tommyblue:
#### :question: What

This is part of the basic tutorial PR https://github.com/AdRoll/baker/pull/85 , adding the example program files to the `main` branch

### DIFF
--- a/filter/clausefilter.go
+++ b/filter/clausefilter.go
@@ -12,7 +12,7 @@ import (
 )
 
 const clauseFilterHelpMsg = `
-This filter lets you set a boolean expression (in s-expression format) that will be matched against all records and dropped if they don't match the expression.
+Discard records which do not match a clause given as a boolean S-expression. Check the filter documentation for some examples.
 
 ### ClauseFilter boolean expression format
 
--- a/filter/clausefilter.go
+++ b/filter/clausefilter.go
@@ -11,21 +11,10 @@ import (
 	"github.com/AdRoll/baker"
 )
 
-// ClauseFilterDesc describes the ClauseFilter filter
-var ClauseFilterDesc = baker.FilterDesc{
-	Name:   "ClauseFilter",
-	New:    NewClauseFilter,
-	Config: &ClauseFilterConfig{},
-	Help:   "This filter lets you set a boolean expression (in s-expression format) that will be matched against all records and dropped if they don't match the expression.\nCheck the filter file for documentation what the format looks like.\n",
-}
+const clauseFilterHelpMsg = `
+This filter lets you set a boolean expression (in s-expression format) that will be matched against all records and dropped if they don't match the expression.
 
-// ClauseFilterConfig describes the ClauseFilter filter config
-type ClauseFilterConfig struct {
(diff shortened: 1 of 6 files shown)


filter/clausefilter.go
```diff
@@ -11,21 +11,10 @@ import (
...
 
-// ClauseFilterDesc describes the ClauseFilter filter
-var ClauseFilterDesc = baker.FilterDesc{
-	Name:   "ClauseFilter",
-	New:    NewClauseFilter,
-	Config: &ClauseFilterConfig{},
-	Help:   "This filter lets you set a boolean expression (in s-expression format) that will be matched against all records and dropped if they don't match the expression.\nCheck the filter file for documentation what the format looks like.\n",
-}
+const clauseFilterHelpMsg = `
+This filter lets you set a boolean expression (in s-expression format) that will be matched against all records and dropped if they don't match the expression.
```
arl, line 15: It's not just on this filter, I think we should do a pass on all docs but IMHO we could avoid starting the doc help string with _this filter_. Also here there's the use of passive. propostion:
```suggestion
Discard records which do not match a clause given as a boolean S-expression. Check the filter documentation for some examples.
```
Since the filter uses a mini lisp-like language to specify the clause, it seems important to ask the user to look at the documentation (or add a link to the doc page on pkg.go.dev -> https://pkg.go.dev/github.com/AdRoll/baker/filter#ClauseFilter)

filter/clear_fields.go
```diff
@@ -13,7 +13,7 @@ var ClearFieldsDesc = baker.FilterDesc{
 	Name:   "ClearFields",
 	New:    NewClearFields,
 	Config: &ClearFieldsConfig{},
-	Help:   "Clear a set of fields of all records passing through",
+	Help:   "Clear a set of fields (makes them empty) of all records passing through",
```
arl, line 16: If clear is not so clear (no pun intended :-) ) maybe we can just use:
```suggestion
	Help:   "Reset a set of fields of all records passing through",
```

tutorials/basic/main.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func fieldByName(name string) (idx baker.FieldIndex, ok bool) {
+	idx, ok = fieldsByName[name]
+	return idx, ok
+}
+
+func fieldName(idx baker.FieldIndex) string {
+	return fields[idx]
+}
```
arl, line 54: As stated, we should probably use the new `[fields]` section here also

arl (approved):
Some minor comments

tutorials/basic/main.go
tommyblue: yup, waiting for the related pr to be merged

filter/clausefilter.go
tommyblue: Ok for rephrasing. I'm not convinced about the "take a look at" because it's the same documentation the user is reading (either the go file or the web page, as they are the same). So the link to the web page will only show the same text but in the browser, if the user is reading the go file (and a link to itself in case of the web page) :thinking:

tutorials/basic/main.go
tommyblue: done

PR #87, opened 2020-11-16
doc: how to test filters

tommyblue:
#### :question: What

how to test filters

### DIFF
--- a/content/en/docs/How Tos/create_filter.md
+++ b/content/en/docs/How Tos/create_filter.md
@@ -217,8 +217,7 @@ When writing tests for a new filter, particular attention should be given to:
 * the New (constructor-like) function 
 * the `Process()` function
 
-Testing the constructor-like function means testing that it is able to intercept wrong
-configurations, i.e. it doesn't return an error.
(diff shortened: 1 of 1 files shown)


content/en/docs/How Tos/create_filter.md
```diff
@@ -209,3 +209,44 @@ func (f *MyFilter) Process(r Record, next func(Record)) {
...
+
+if filter, err := NewMyFilter(cfg); err != nil {
+    t.Errorf("unexpected error: %v", err)
+}
+```
+
+> Obviously, if the filter requires some configuration values (not like this empty demo filter),
+the test should also test all possible values and corner cases.
+
+With the `filter` instance, is then possible to test the `Process()` function, providing a
```
arl, line 240:
```suggestion
With the `filter` instance, it's then possible to test the `Process()` function, providing a
```

arl (approved):
Some comments/propositions.

For how-tos, guides or tutorials kind of docs, I like the user of _we_, to involve the reader, like you're showing them the way to do stuff, with them following along. But it's probably a matter of personal taste, just wanted to give my 2c about that :-)

PR #89, opened 2020-11-17
doc: how to test uploads

tommyblue:
#### :question: What

how to test uploads

### DIFF
--- a/content/en/docs/How Tos/create_upload.md
+++ b/content/en/docs/How Tos/create_upload.md
@@ -79,7 +79,9 @@ wisely and documented extensively.
 
 ## Write tests
 
-Testing an upload component can seem sometimes a bit useless as it probably uses external resources
+Since by definition an upload component involves external resources you either have to mock that resource 
+or use it directly. See an example of how to mock an external resource in the [insert s3 mock test link]. 
+However writing a test that uses the actual external resource (a.k.a end-to-end testing) is out of the scope of this how-to.
 and libraries for most of its job.
 
 We thereby provide some suggestions to test those components:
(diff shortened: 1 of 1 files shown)


content/en/docs/How Tos/create_upload.md
```diff
@@ -76,3 +76,21 @@ Through the channel, the upload receives from the output paths to local files th
 The only Upload component implemented at the moment, S3, removes those files once uploaded, but there isn't a
 golden rule for what to do with them. This is up to the upload component and should be chosen
 wisely and documented extensively.
+
+## Write tests
+
+Testing an upload component can seem sometimes a bit useless as it probably uses external resources
```
tommyblue, line 82: I'm not 100% sure of this sentence. It's what I mean, but maybe we could write it in another way. @arl any suggestion?

arl: I agree, I wouldn't say it like that.
```suggestion
Since by definition an upload component involves external resources you either have to mock that resource 
or use it directly. See an example of how to mock an external resource in the [insert s3 mock test link]. 
However writing a test that uses the actual external resource (a.k.a end-to-end testing) is out of the scope of this how-to.
```

content/en/docs/How Tos/create_upload.md
```diff
@@ -76,3 +76,21 @@ Through the channel, the upload receives from the output paths to local files th
...
+and libraries for most of its job.
+
+We thereby provide some suggestions to test those components:
+
+* do not test external libraries when possible, they should be already tested in their packages
```
arl, line 87: true for unit-tests, but writing an end-to-end test is always a nice thing though

arl (approved):
See suggestion for your question.
If you go with it, you may have to reformulate a bit (or remove) some parts written below it in order to avoid duplication.

Otherwise it LGTM

PR #90, opened 2020-11-17
doc: how to test inputs

tommyblue:
#### :question: What

how to test inputs

### DIFF
--- a/content/en/docs/How Tos/create_input.md
+++ b/content/en/docs/How Tos/create_input.md
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
 All these parameters appear in the generated help. `help` should describe the parameter role and/or
 its possible values, `required` informs Baker it should refuse configurations in which that field
 is not defined.
+
+## Write tests
+
+To test an input component we suggest two main paths:
+
+* test the component in isolation, calling the `Run` function
+* test the input at high-level, running a complete Baker topology
+
+Regardless of the chosen path, two additional unit tests are always suggested:
+
+* test the `New()` (constructor-like) function, to check that the function is able to correctly
+instantiate the component with valid configurations and intercept wrong ones
+* create small and isolated functions where possible and unit-test them
+
+### Test calling Run()
+
+In case we want to test the component calling the `Run` function, this is an example of test where,
+after some initialization, the `input.Run` function is called and the produced `Data` is checked
+in a goroutine:
+
+```go
+func TestMyInput(t *testing.T) {
+    ch := make(chan *baker.Data)
+    defer close(ch)
+
+    go func() {
+        for data := range ch {
+            // test `data`, that comes from the component,
+            // like checking its content, parse the records, metadata, etc
+            if something_is_wrong(data) {
+                t.Fatalf("error!")
+            }
+        }
+    }()
+
+    // configure the input
+    cfg := ...
+
+    input, err := NewMyInput(cfg) // use the contructor-like New function
+    // check err
+
+    // if the input requires other things, initialize/create them
+
+    // run the input
+    if err := input.Run(ch); err != nil {
+        t.Fatal(err)
+    }
+}
+```
+
+The `List` input [has an example](https://github.com/AdRoll/baker/blob/main/input/list_test.go)
+of this testing strategy.
+
+### Test the component running a topology
+
+If we want to test the component creating and running a topology, we need to create one starting
+from the TOML configuration and then calling `NewConfigFromToml`, `NewTopologyFromConfig` and `Run`.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
 All these parameters appear in the generated help. `help` should describe the parameter role and/or
 its possible values, `required` informs Baker it should refuse configurations in which that field
 is not defined.
+
+## Write tests
+
+To test an input component we suggest two main paths:
+
+* test the component in isolation, calling the `Run` function
+* test the input at high-level, running a complete Baker topology
```
arl, line 104:
```suggestion
* write an higher-level test by running a complete Baker topology
```

content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
...
+
+Regardless of the chosen path, two additional unit tests are always suggested:
+
+* test the `New()` (constructor-like) function, to check that the function is able to correctly
+instantiate the component with valid configurations and intercept wrong ones
```
arl, line 109:
```suggestion
instantiate the component with valid configurations and intercept incorrect ones (in case that's possible)
```

content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
...
+        for data := range ch {
+            // test `data`, that comes from the component,
+            // like checking its content, parse the records, metadata, etc
+            if something_is_wrong(data) {
+                t.Fatalf("error!")
+            }
+        }
+    }()
+
+    // configure the input
```
arl, line 133: nit: Caps and punctuation! :rofl: 
```suggestion
    // Configure the input.
```

content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
...
+    cfg := ...
+
+    input, err := NewMyInput(cfg) // use the contructor-like New function
+    // check err
+
+    // if the input requires other things, initialize/create them
```
arl, line 139: Not sure I'd write that. Either we find a more specific example where it's needed or we shouldn't write it IMHO

content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
...
+In case we want to test the component calling the `Run` function, this is an example of test where,
+after some initialization, the `input.Run` function is called and the produced `Data` is checked
+in a goroutine:
+
+```go
+func TestMyInput(t *testing.T) {
+    ch := make(chan *baker.Data)
+    defer close(ch)
+
+    go func() {
```
arl, line 123: Maybe we could add a comment saying we start a goroutine that acts as Baker, consuming the baker.Data produced by our tested input?

content/en/docs/How Tos/create_input.md
```diff
@@ -95,3 +95,117 @@ Each field in the struct must include a `help` string tag (mandatory) and a `req
...
+
+    // run the input
+    if err := input.Run(ch); err != nil {
+        t.Fatal(err)
+    }
+}
+```
+
+The `List` input [has an example](https://github.com/AdRoll/baker/blob/main/input/list_test.go)
+of this testing strategy.
```
arl, line 149: The List input unit-tests are a bit complex, hard to follow, I'd either link to a specific test in list_test.go rather than the whole file.
Or, but it would take way more time, so if you agree we could add a low-priority ticket for that, write a test for the kinesis input in which kinesis is mocked, and link to that test, as it should be simpler to follow. Currently kinesis input is not tested unfortunately.
Or if you find any other input unit-test that is simpler, that would be nice, if that exists

arl (approved):
Some small suggestions and nits, as always :D

content/en/docs/How Tos/create_input.md
tommyblue: ok, I linked a specific test but I agree about writing an ad-hoc test. I'll create a ticket for that

content/en/docs/How Tos/create_input.md
tommyblue: yup, adding

content/en/docs/How Tos/create_input.md
tommyblue: right, removing

PR #92, opened 2020-11-20
Website home

tommyblue:
#### :question: What

rework the home page


content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,33 @@
...
+title: "Performance"
+weight: 150
+date: 2020-11-20
+description: >
+  Baker performance test cases
+---
+
+### Read from S3 and write to local disk
+
+On a `c5.2xlarge` instance, Baker managed to read zstandard records from S3, uncompress them and
```
arl, line 11: Not sure c5.2xlarge speaks to everybody, maybe we could add in parenthesis (or as a footer note as you prefer) that this is an AWS EC2 instance based on Intel Xeon Platinum 8000. Or maybe just that is an AWS instance type.
```suggestion
On a `c5.2xlarge` instance, Baker can read zstandard records from S3, uncompress them and
```

arl (commented):
Performance:
zstd is not a compression format which is as popular as others, mostly because it's quite recent so I'm not sure it's going to speak to everybody when they see `.zstd`.
About the numbers, I know 178k records per second is a lot, but I know it since I've got an idea about the records we generally process at Adroll, but 178k records per second won't really mean anything for somebody else. Counter-example: if my records are 64b long, that's not very impressive :grimacing: . I guess what I'm trying to say is there's a unit which would speak to everybody, the throughput, in MB/s or GB/s, also this number should be associated with some context. Maybe we can add one line of context taken from the performance page?

In the High-performance drawing, it seems that the 3 texts below correspond to the 3 icons above but they're not, right?

content/en/_index.html
```diff
@@ -3,108 +3,56 @@
...
-	var duration time.Duration
-	err = baker.Main(cfg, duration)
-	if err != nil {
-		log.Fatal(err)
-	}
-}
-{{< / highlight >}}
-</div>
+{{% blocks/feature icon="fas fa-puzzle-piece" title="Extendable" %}}
+With Baker it's super easy to write new components or process any kind of structured data.
```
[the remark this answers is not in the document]
tommyblue: ok

content/en/_index.html
```diff
@@ -3,108 +3,56 @@
...
-{{% /blocks/feature %}}
+{{< blocks/section color="primary" >}}
+{{% blocks/feature icon="fas fa-link" title="Modular pipeline" %}}
 
-{{% blocks/feature icon="fab fa-github" title="Contributions welcome!" %}}
-We encourage all users to contribute to the project by adding new components, improve the core or the documentation.
+Baker uses modular components in a 4-levels stack: inputs, filters, outputs and uploads.
 
-Visit the project on GitHub and feel free to open a new [Pull Request](https://github.com/AdRoll/baker/pulls).
+Each component of a level can connect with any other from the previous and the next one.
```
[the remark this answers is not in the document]
tommyblue: rephrased a bit

content/en/_index.html
```diff
@@ -3,108 +3,56 @@
 linkTitle = "Baker"
 +++
 
-{{< blocks/cover title="High performance, composable and extendable data-processing pipeline for the big data era" image_anchor="top" height="med" color="blue" >}}
+{{< blocks/cover title="High performance modular pipelines for the Big Data era" image_anchor="top" height="min" color="blue" >}}
```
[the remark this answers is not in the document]
tommyblue: I've changed it a bit, setting Baker as title and the text as subtitle

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+
+Within NextRoll, Baker is often executed on AWS EC2 instances, and thus you find in this page
+many references to
+[EC2 instance types](https://aws.amazon.com/ec2/instance-types/) (`c5.2xlarge`, `c5.2xlarge`, etc.).
+
+### Read from S3 and write to local disk
+
+On an AWS EC2 instance of size `c5.2xlarge`, Baker can read zstandard records from S3, uncompress
+them and apply a basic filtering logic, compressing them back on local files with zstandard at
+compression level 3 and long range mode at 27 using ~90% of capacity of each vCPU (8 in total) and
```
arl, line 20: I wouldn't write long range mode without explaining what that means here since it's very technical and very specific.
But I'm not sure we should write that, nor compression level 3 for that matter.

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+~3.5GB of RAM.  
+
+It reads and writes a total of 94 million records in 8'51" (~178k r/w records per second).
```
arl, line 23: Since we already wrote that baker reads and writes, we should probably just stick to `178k records per second`.
Again, we should round things up a bit in order to make it easily readable IMHO 
```suggestion
It reads and writes a total of 94 million records in less than 9 minutes, that's 178k records per second.
```

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+
+On a `c5.2xlarge` instance (48vCPUs) the same test takes 2'2" (~775k r/w records per second).
```
arl, line 25: Same here, the `'` `''` notation doesn't make for an easy read IMO
```suggestion
On a `c5.2xlarge` instance (48 vCPUs) the same test takes 2 minutes, so that's a speed of 775k records per second.
```

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+
+For this test we use 711 zstd compressed files for a total of 17 GB of compressed size and 374 GB
+of uncompressed size. The average size of each record is ~4.5 KB.
```
arl, line 28: Since we already say it's an average then the `~` symbol can be removed
```suggestion
For this test we use 711 zstd compressed files for a total of 17 GB of compressed size and 374 GB
of uncompressed size. The average size of each record is 4.5 KB.
```

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Read from S3 and write to DynamoDB (in the same region)
+
+On a `c5.4xlarge` instance, Baker reads zstd compressed files from S3 writing to DynamoDB (configured
+with 20k write capacity units) at an average speed of 60k records/s (the average size of each record
+is 4.3 KB) using less than 1 GB of memory and ~300% of the total CPU capacity (less than 20% for
```
arl, line 34: This suggestion just removes `~`
```suggestion
is 4.3 KB) using less than 1 GB of memory and 300% of the total CPU capacity (less than 20% for
```

Not sure what _less than 20%_ means? _less than 20% of CPU usage_ maybe?

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+each core).
+
+The bottleneck here is the DynamoDB write capacity, so Baker can easily cope with an increased load
+just increasing the write capacity units in DynamoDB (up to 400k).
```
arl, line 38: repetition of increased.
repetition of dynamodb
```suggestion
The bottleneck here is the DynamoDB write capacity, so Baker could handle the additional load
caused by increasing the write capacity, up to 400k.
```
I dont understand where does that 400k come from? if at 20k write capacity we use 20% of CPU, that means that we would be maxed out at 100% CPU if we multiply the write capacity by 5, so 100k write capacity.
What am i missing?

content/en/docs/Performance/_index.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Read from Kinesis and write to DynamoDB (in the same region)
+
+On a `c5.4xlarge` instance, we performed a test reading from a Kinesis stream with 130 shards and
+writing to a DynamoDB table with 20k write capacity units. Baker is able to read and write more
+than 10k records per second (the avg size of each record is 4.5 KB) using less than 1 GB of RAM and
+~400% of the total CPU capacity (less than 25% for each core).
```
arl, line 45:
```suggestion
than 10k records per second (the average size of each record is 4.5 KB) using less than 1 GB of RAM and
around 400% of the total CPU capacity (each core being used at less than 25%).
```

arl (commented):
I let a lot of comments on the performance page, in order to make it more easily palatable, which is always a bit hard to do with numbers.

content/en/docs/Performance/_index.md
tommyblue: ok, the number is the max write capacity you can set in dynamodb, not considering our cpu. We can clarify with `just increasing the write capacity units in DynamoDB (whose max value can get up to 400k).` or similar

arl: Oh ok so I'm not sure It's super relevant for this page then. We can just say baker would be able to handle an additional load, without writing that number

tommyblue:
```suggestion
The bottleneck here is the DynamoDB write capacity, so Baker could handle the additional load caused by
a possible increase in the write capacity.
```

content/en/docs/Performance/_index.md
tommyblue: removed

arl (approved):
LGTM

PR #96, opened 2020-11-25
input: sqs: fix send on closed channel

arl:
#### :question: What

Fix the `panic: send on closed channel` happening very regularly when stopping a running pipeline with the SQS input with CTRL-C.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/input/sqs.go
+++ b/input/sqs.go
@@ -1,6 +1,7 @@
 package input
 
 import (
+	"context"
 	"encoding/json"
 	"net/url"
 	"regexp"
@@ -52,11 +53,13 @@ func (cfg *SQSConfig) fillDefaults() {
 }
 
 type SQS struct {
-	*inpututils.S3Input
+	s3Input *inpututils.S3Input
 
 	Cfg            *SQSConfig
 	FilePathRegexp *regexp.Regexp
 	svc            *sqs.SQS
+	wg             sync.WaitGroup
+	done           chan bool
 
 	minSnsTimestamp time.Time
 }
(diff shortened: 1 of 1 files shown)


input/sqs.go
```diff
@@ -216,25 +231,38 @@ func (s *SQS) Run(inch chan<- *baker.Data) error {
 		for _, url := range resp.QueueUrls {
 			wg.Add(1)
 			go func(url string) {
-				defer wg.Done()
-				s.pollQueue(url)
+				defer func() {
```
tommyblue, line 234: nit: can it be one-lined?

input/sqs.go
```diff
@@ -52,11 +53,13 @@ func (cfg *SQSConfig) fillDefaults() {
 }
 
 type SQS struct {
-	*inpututils.S3Input
+	s3Input *inpututils.S3Input
```
tommyblue, line 56: out of curiosity: is there a reason for removing the embedding? Style? Other? thx

input/sqs.go
```diff
@@ -216,25 +231,38 @@ func (s *SQS) Run(inch chan<- *baker.Data) error {
...
+	//  - first we close the 'done' channel, this in turns cancel polling via
+	//    context cancellation
+	//  - we then wait for all the polling goroutines to end.
+	//  - After this point we are guaranteed to not receive any more files so
+	//    we notify the embedded S3input.
+	//  - now we ask S3Input to stop as soon as it has finished processing files
+	//  - finally Run exits after being signaled from S3Input that we can
+	<-s.done
+	cancel()
+	wg.Wait()
```
tommyblue, line 253: what do you think about adding a test for this stop procedure?

tommyblue (approved):
LGTM, I'd maybe add a test for the stop steps

input/sqs.go
arl: Since `inpututils.S3Input` already implements the `baker.Input` interface, `SQS` didn't have to implement the `Stop` method to also implement it since it anonymously embbeded `S3Input`., and when the topology called `SQS.Stop`, this was in fact `S3Input.Stop`that was called.

We now do want to provide our own implementation of `Stop`, and though we could have done it without having this change I do believe it's better to leave no doubt to the reader as to which Stop method is called, that why I chose to un-embed it

input/sqs.go
arl: done

input/sqs.go
arl: I agree about adding a test, maybe in another PR

PR #104, opened 2020-12-03
Link docs to api

tommyblue:
#### :question: What

add a header to each components page linking the API section for the component family

### DIFF
--- a/content/en/docs/Components/Filters/ClauseFilter.md
+++ b/content/en/docs/Components/Filters/ClauseFilter.md
@@ -3,6 +3,11 @@ title: "ClauseFilter"
 weight: 8
 date: 2020-12-03
 ---
+{{% pageinfo color="primary" %}}
+
+**Read the [API documentation &raquo;](https://pkg.go.dev/github.com/AdRoll/baker/filter)**
+{{% /pageinfo %}}
+
 ## Filter *ClauseFilter*
 
 ### Overview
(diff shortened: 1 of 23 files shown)


content/en/docs/Components/Filters/ClauseFilter.md
```diff
@@ -3,6 +3,11 @@ title: "ClauseFilter"
 weight: 8
 date: 2020-12-03
 ---
+{{% pageinfo color="primary" %}}
+
+**Read the [API documentation &raquo;](https://pkg.go.dev/github.com/AdRoll/baker/filter)**
```
arl, line 8: This links to https://pkg.go.dev/github.com/AdRoll/baker/filter rather than to https://pkg.go.dev/github.com/AdRoll/baker/filter#ClauseFilter and same for all components, sorry... :-)

Also, not related to your PR, but if you look at https://github.com/AdRoll/baker/blob/4042d80bf9bae06adce1dcf3c21c862c792b0c7d/filter/clausefilter.go#L14-L69
it's too bad this documentation only appears in the code and nowhere else. We should probably copy-paste it into the component struct GoDoc. It's generally true for most, if not all components, they don't have Godoc, but they should....

tommyblue: fixed the links in this PR, but it will require additional commits after https://github.com/AdRoll/baker/pull/105 has been merged

arl (approved):

content/en/docs/Components/Filters/ClauseFilter.md
arl: @tommyblue tell me if you intend to do it or I'll create a ticket

PR #106, opened 2020-12-03
TimestampRange filter accepts 'now' as bound

tommyblue:
#### :question: What

`TimestampRange` filter accepts `"now"` as time bound (both start or end)


filter/timestamp_range_test.go
```diff
@@ -19,14 +20,14 @@ func TestTimestampRange(t *testing.T) {
 			name:      "valid",
 			ts:        1580981641, // 2020-02-06 09:34:01
 			startDate: "2019-02-06 09:34:01",
-			endDate:   "2022-02-06 09:34:01",
+			endDate:   "2122-02-06 09:34:01",
```
arl, line 23: Why is this change necessary?

filter/timestamp_range_test.go
```diff
@@ -19,14 +20,14 @@ func TestTimestampRange(t *testing.T) {
...
 			want:      true,
 		},
 		{
 			name:      "past lower bound",
 			ts:        1528277641, // 2018-06-06 09:34:01
 			startDate: "2019-02-06 09:34:01",
-			endDate:   "2022-02-06 09:34:01",
+			endDate:   "2122-02-06 09:34:01",
```
arl, line 30: same question

filter/timestamp_range_test.go
```diff
@@ -50,6 +51,20 @@ func TestTimestampRange(t *testing.T) {
...
+			want:      true,
+		},
+		{
+			name:      "end now",
+			ts:        int(time.Now().Unix()),
+			startDate: "2018-02-06 09:34:01",
+			endDate:   "now",
+			want:      false,
+		},
```
arl, line 67: would be nice to see a test with `"now"` for both bounds

filter/timestamp_range.go
```diff
@@ -68,6 +66,15 @@ func NewTimestampRange(cfg baker.FilterParams) (baker.Filter, error) {
...
 
+func (c *TimestampRangeConfig) getTime(s string) (time.Time, error) {
+	if s == "now" {
+		return time.Now(), nil
+	}
+
+	const timeLayout = "2006-01-02 15:04:05"
+	return time.Parse(timeLayout, s)
+}
+
```
arl, line 77: Not sure this addition warrants a new function IMHO.
I'm not against a tiny bit of repetition, when it helps readability and doesn't hurt performance where you'd need some.

Also `getTime` has been added on the config struct, so from a readability standpoint it shouldn't be declared in-between `NewTimestampRange` and the `Stats` method.

Why not do things the other way around and call the function `setTimes(start, end string) error`?

arl (approved):
Nice addition.
Some comments about readability

filter/timestamp_range_test.go
tommyblue: it's not, I thought about "now" I then I updated to a date more in the future, but I realize we use `ts` for the test

filter/timestamp_range_test.go
tommyblue: done

arl (approved):

PR #107, opened 2020-12-04
add PartialClone filter

tommyblue:
#### :question: What

Add a new filter that creates a new record copying a list of fields and sends this new record through the filter chain


filter/partial_clone.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if len(dcfg.Fields) == 0 {
+		return nil, fmt.Errorf("PartialClone: add at least one field")
+	}
+
+	fieldIdx := []baker.FieldIndex{}
+	for _, f := range dcfg.Fields {
+		idx, ok := cfg.FieldByName(f)
+		if !ok {
+			return nil, fmt.Errorf("Can't resolve field name %s", f)
```
arl, line 43: nit: it's better for an error message to not start with caps in order to remain syntactically correct if it was to be wrapped with another before printed to the user

filter/partial_clone_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	tests := []struct {
+		name       string
+		record     string
+		wantRecord string
+	}{
+		{
+			name:       "base",
+			record:     "a,foo,bar",
+			wantRecord: "a,,bar,,",
```
arl, line 104: I think it would probably be better if we weren't comparing the serialized version of the processed record. These trailing `,` are due to a bug that still needs to be fixed.
Why not call a serie of record.Get and Set on it?

filter/partial_clone_test.go
```diff
@@ -0,0 +1,139 @@
...
+		case "field1":
+			return 1, true
+		case "field2":
+			return 2, true
+		default:
+			return -1, false
+		}
+	}
+
+	tests := []struct {
```
arl, line 96: no need for a slice of tests if there's just a single case

arl (commented):
I think both tests could easily be merged into a single one that:
 - first checks that the config gives an error (or not),
 - and goes on (only if wantErr is false) by applying that config to a record and testing the outcome

example in https://github.com/AdRoll/baker/blob/main/filter/replace_fields_test.go

filter/partial_clone_test.go
tommyblue: the idea was to avoid to call extra functions (to avoid adding more hidden logic), but I see the point here. fixed

arl (approved):

PR #109, opened 2020-12-09
Update Pipeline Configuration with missing confs

tommyblue:
#### :question: What

we missed a bunch of conf sections

### DIFF
--- a/content/en/docs/How Tos/pipeline_configuration.md
+++ b/content/en/docs/How Tos/pipeline_configuration.md
@@ -9,10 +9,10 @@ description: >
 A Baker [pipeline](/docs/core-concepts/#pipeline) is declared in a configuration
 file in [TOML format](https://toml.io/en/).
 We use this file to:
- * define the topology (i.e the list of components) of the pipeline we want to run
- * configure each component
- * setup general elements such as metrics
 
+* define the topology (i.e the list of components) of the pipeline we want to run
+* configure each component
+* setup general elements such as metrics
 
 ### Configuration file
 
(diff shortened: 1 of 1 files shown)


arl:
Not sure about documenting the `[csv]` section since ideally it should go as it's specific to a particular Record implementation, which might not even be present if baker is used as a library.
The rest LGTM

tommyblue:
> Not sure about documenting the `[csv]` section since ideally it should go as it's specific to a particular Record implementation, which might not even be present if baker is used as a library.

I see, but it's also true that the docs should document what's in the project. I think we could add it (as it's present now) and we'll remove once removed from the code

arl:
> > Not sure about documenting the `[csv]` section since ideally it should go as it's specific to a particular Record implementation, which might not even be present if baker is used as a library.
> 
> I see, but it's also true that the docs should document what's in the project. I think we could add it (as it's present now) and we'll remove once removed from the code

The thing is that once you're documenting something, it's in theory harder to remove since you influence people to use it.

tommyblue:
> The thing is that once you're documenting something, it's in theory harder to remove since you influence people to use it.

hehe, we already have a PR on that :smile: 
So let's choose quickly what to do then

content/en/docs/How Tos/pipeline_configuration.md
```diff
@@ -83,6 +103,10 @@ can be specified, and in this case we simply provide the files option to List.
 Notice that List would accept http:// or even s3:// URLs there in addition to local paths,  
 and some more (run ./Baker-bin -help List in the help example for more details).
 
+`[filterchain]` defines the configuration for the whole filter chain. Filter-specific configuration
+are provided by `[[filter]]` (see below). The only accepted configuration is `procs = <int>` that
```
arl, line 107: Either:
 - Filter-specific configuration is provided
 - Filter-specific configurations are provided

I guess

arl (approved):
Thanks
LGTM

There's just a grammar comment

PR #110, opened 2020-12-09
Add lua filter

arl:
#### :question: What

Add a LUA filter that allows to add filter logic (implement the Process function) in lua.
Needs further testing/benchmarking/polishing but first results shows a relatively good performance with respect to the same filter coded in Go (5% slower). 

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/filter/lua.go
+++ b/filter/lua.go
@@ -22,8 +22,8 @@ type LUAConfig struct {
 }
 
 type LUA struct {
-	l        *lua.LState
-	funcName string
+	l       *lua.LState
+	luaFunc lua.LValue
 }
 
 func NewLUA(cfg baker.FilterParams) (baker.Filter, error) {
@@ -34,10 +34,12 @@ func NewLUA(cfg baker.FilterParams) (baker.Filter, error) {
 		return nil, fmt.Errorf("can't compile lua script %q: %v", dcfg.Script, err)
 	}
 	registerLUARecordType(l)
+	// TODO: check function exists
+	luaFunc := l.GetGlobal(dcfg.FilterName)
 
 	f := &LUA{
-		funcName: dcfg.FilterName,
-		l:        l,
+		luaFunc: luaFunc,
+		l:       l,
 	}
 
 	runtime.SetFinalizer(f, func(f *LUA) { f.l.Close() })
@@ -49,13 +51,13 @@ func (t *LUA) Stats() baker.FilterStats { return baker.FilterStats{} }
 
 func (t *LUA) Process(rec baker.Record, next func(baker.Record)) {
 	luaNext := t.l.NewFunction(func(L *lua.LState) int {
-		recordArg := checkLuaRecord(L, 1)
+		recordArg := fastcheckLuaRecord(L, 1)
 		next(recordArg.r)
 		return 0
 	})
 
 	err := t.l.CallByParam(lua.P{
-		Fn:      t.l.GetGlobal(t.funcName),
+		Fn:      t.luaFunc,
 		NRet:    0,
 		Protect: true,
 	}, recordToLua(t.l, rec),
@@ -99,9 +101,13 @@ func checkLuaRecord(L *lua.LState, n int) *luaRecord {
 	return nil
 }
 
+func fastcheckLuaRecord(L *lua.LState, n int) *luaRecord {
+	return L.Get(n).(*lua.LUserData).Value.(*luaRecord)
+}
+
 // record:get(int) returns string
 func luaRecordGet(L *lua.LState) int {
-	luar := checkLuaRecord(L, 1)
+	luar := fastcheckLuaRecord(L, 1)
 	fidx := L.CheckInt(2)
 
 	buf := luar.r.Get(baker.FieldIndex(fidx))
@@ -112,7 +118,7 @@ func luaRecordGet(L *lua.LState) int {
 
 // record:set(int, string)
 func luaRecordSet(L *lua.LState) int {
-	luar := checkLuaRecord(L, 1)
+	luar := fastcheckLuaRecord(L, 1)
 	fidx := L.CheckInt(2)
 	val := L.CheckString(3)
 
--- a/filter/lua.go
+++ b/filter/lua.go
@@ -1,0 +1,122 @@
+package filter
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/AdRoll/baker"
+	lua "github.com/yuin/gopher-lua"
+)
+
+// LUADesc describes the LUA filter
+var LUADesc = baker.FilterDesc{
+	Name:   "LUA",
+	New:    NewLUA,
+	Config: &LUAConfig{},
+	Help:   `TBD`,
+}
+
+type LUAConfig struct {
+	Script     string
+	FilterName string
+}
+
+type LUA struct {
+	l        *lua.LState
+	funcName string
+}
+
+func NewLUA(cfg baker.FilterParams) (baker.Filter, error) {
+	dcfg := cfg.DecodedConfig.(*LUAConfig)
+
+	l := lua.NewState()
+	if err := l.DoFile(dcfg.Script); err != nil {
+		return nil, fmt.Errorf("can't compile lua script %q: %v", dcfg.Script, err)
+	}
+	registerLUARecordType(l)
+
+	f := &LUA{
+		funcName: dcfg.FilterName,
+		l:        l,
+	}
+
+	runtime.SetFinalizer(f, func(f *LUA) { f.l.Close() })
+
+	return f, nil
+}
+
+func (t *LUA) Stats() baker.FilterStats { return baker.FilterStats{} }
+
+func (t *LUA) Process(rec baker.Record, next func(baker.Record)) {
+	luaNext := t.l.NewFunction(func(L *lua.LState) int {
+		recordArg := checkLuaRecord(L, 1)
+		next(recordArg.r)
+		return 0
+	})
+
+	err := t.l.CallByParam(lua.P{
+		Fn:      t.l.GetGlobal(t.funcName),
+		NRet:    0,
+		Protect: true,
+	}, recordToLua(t.l, rec),
+		luaNext)
+
+	if err != nil {
+		panic(err)
+	}
+}
+
+const luaRecordTypeName = "record"
+
+func registerLUARecordType(L *lua.LState) {
+	mt := L.NewTypeMetatable(luaRecordTypeName)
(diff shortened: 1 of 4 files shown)


lucarin91 (changes requested):

filter/lua.go
```diff
@@ -0,0 +1,342 @@
...
+	fields := l.NewTable()
+	for fidx, fname := range comp.FieldNames {
+		fields.RawSet(lua.LNumber(fidx), lua.LString(fname))
+	}
+	l.SetGlobal("fieldNames", fields)
+}
+
+// TODO: at the moment LUA filter doesn't publish stats.
+// There are multiple ways to do it, either require the filter to update
+// the numbers of processed and filtered records, or deduce them automatically
```
lucarin91, line 225: I think that this second hypothesis could be the best one. We could count how many times the next function is called, and update the `numProcessedLines` and `numFilteredLines` atomic accordingly.

filter/lua.go
```diff
@@ -0,0 +1,342 @@
...
+- Go: ` + ticks("Components.FieldNames []string") + `
+- lua: ` + ticks("fieldNames") + `
+
+` + ticks("fieldNames") + ` is an integed-indexed table, in other words an array, containing all field names, as ` + ticks("FieldNames") + ` in Go.
+`
+
+// TODO(arl): ideally this functions should not be required, but writing
+// markdown documentation in Go strings is tedious and error-prone.
+
+func ticks(s string) string         { return "`" + s + "`" }
```
tommyblue, line 109: I'm not sure I see the advantage of this function as you need to call it with 
```
` + ticks("string") + `
```
The use of
```
\`string\`
``` 
seems a bit better to me

filter/lua.go
```diff
@@ -0,0 +1,353 @@
...
+It allows to lookup a field index by its name, returning the index or nil if no field exists with this index.
+
+### Global tables
+
+#### fieldNames
+
+- Go: ` + ticks("Components.FieldNames []string") + `
+- lua: ` + ticks("fieldNames") + `
+
+` + ticks("fieldNames") + ` is an integed-indexed table, in other words an array, containing all field names, as ` + ticks("FieldNames") + ` in Go.
```
tommyblue, line 103:
```suggestion
` + ticks("fieldNames") + ` is an integer-indexed table, in other words an array, containing all field names, as ` + ticks("FieldNames") + ` in Go.
```

filter/lua.go
```diff
@@ -0,0 +1,342 @@
...
+
+	registerLUATypes(l, cfg.ComponentParams)
+
+	luaFunc := l.GetGlobal(dcfg.FilterName)
+	if luaFunc.Type() == lua.LTNil {
+		return nil, fmt.Errorf("can't find lua filter %q in script %q", dcfg.FilterName, dcfg.Script)
+	}
+
+	// Preallocate the userdata we use to wrap the record passed to the filter.
+	// We can do this since a single instance of a baker filter is only ever
```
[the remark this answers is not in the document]
arl: You're right good catch. The filter was made to only work with a non-concurrent filter chain, which could be ok. But since the filter has no way to know the current value of the `[filterchain] procs` (and it's better in this way) I made the necessary changes to make the filter work in any situation. In other words I removed the pre-allocations and protected the lua state between concurrent calls to `Process`.

filter/lua.go
```diff
@@ -0,0 +1,342 @@
...
+
+// LUAConfig holds the configuration for the LUA filter.
+type LUAConfig struct {
+	Script     string `help:"Path to the lua script where the baker filter is defined" required:"true"`
+	FilterName string `help:"Name of the lua function to run as baker filter" required:"true"`
+}
+
+// LUA allows to run a baker filter from an external script written in lua.
+type LUA struct {
+	l       *lua.LState    // lua state used during all the baker filter lifetime
```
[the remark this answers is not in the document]
arl: see comment above. :+1:

filter/lua.go
```diff
@@ -0,0 +1,342 @@
...
+	err := t.l.CallByParam(lua.P{
+		Fn:      t.luaFunc,
+		NRet:    0,
+		Protect: true,
+	}, t.ud, t.luaNext)
+
+	// TODO: should not panic here and instead increment a filter-specific
+	// metric that tracks the number of lua runtime errors.
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
arl: I thought about that but as I see it the lua call returned an error that means that something very wrong happened inside the filter. In general I'd consider this kind of errors as _programmer errors_ (in lua) and I'm not sure it's worth going ahead with the topology. Maybe we could make it configurable later, for example `ExitOnLuaError bool` or something like that.

filter/lua.go
arl: Since we don't have the optimizations anymore, implementing stats has become trivial now. it's implemented

filter/lua.go
```diff
@@ -0,0 +1,353 @@
...
+	lua "github.com/yuin/gopher-lua"
+)
+
+var luaHelp = `
+This filter runs a baker filter defined in a LUA script.
+It's useful to quickly write and run a Baker filter without having to recompile Baker.
+
+This filter is based on [GopherLua](github.com/yuin/gopher-lua), which is an LUA5.1 virtual machine.
+
+To use this filter you need to declare a function in an lua file. This function serves the same purpose
```
tommyblue, line 20: maybe it's worth telling the user that the function name must be the same as `FilterName` conf

filter/lua.go
```diff
@@ -0,0 +1,353 @@
...
+	l := lua.NewState()
+	if err := l.DoFile(dcfg.Script); err != nil {
+		return nil, fmt.Errorf("can't compile lua script %q: %v", dcfg.Script, err)
+	}
+
+	registerLUATypes(l, cfg.ComponentParams)
+
+	luaFunc := l.GetGlobal(dcfg.FilterName)
+	if luaFunc.Type() == lua.LTNil {
+		return nil, fmt.Errorf("can't find lua filter %q in script %q", dcfg.FilterName, dcfg.Script)
```
tommyblue, line 153:
```suggestion
		return nil, fmt.Errorf("can't find lua filter function name %q in script %q", dcfg.FilterName, dcfg.Script)
```

filter/lua.go
```diff
@@ -0,0 +1,353 @@
...
+		NumProcessedLines: atomic.LoadInt64(&t.nprocessed),
+		NumFilteredLines:  atomic.LoadInt64(&t.nfiltered),
+	}
+}
+
+// Process forwards records to the lua-written filter.
+func (t *LUA) Process(rec baker.Record, next func(baker.Record)) {
+	atomic.AddInt64(&t.nprocessed, 1)
+
+	t.mu.Lock()
```
lucarin91, line 229: Are you sure to lock all the Process functions? This means that even if the Lua function is pure it will not exploits any concurrent execution on different disjoined records. Moreover, it will imply that adding a single LuaFilter on a possibly long and complex filter-chain will inhibit any parallel execution, blocking all the execution on a single lock.
IHMO we need to create a new state each time to avoid the lock.

filter/lua.go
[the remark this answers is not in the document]
lucarin91: Yes, you are right

filter/lua.go
arl: this doesn't work inside a backtick escaped string.
https://play.golang.org/p/HGny5sW5qjW

lucarin91: I think that is not possible to escape back-ticks in a multi-line go string [ref](https://stackoverflow.com/questions/21198980/how-to-escape-back-ticks)

filter/lua.go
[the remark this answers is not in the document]
tommyblue: wrt the proposal to create a new user structure / state for each record, it's probably worth measure the impact of the operation with a huge amount of records (as we usually have)

filter/lua.go
tommyblue: unlocking before `next` would solve the filterchain parallelization issue, no?

arl: I think we should refocus the review on correctness and semantics of the user-api. 
Creating a new state each time is probably quite expensive, it might be beneficial in some cases, in might not be in others and I'm not sure one can be certain it's the way to go to obtain the best overall performance in a majority of cases. 

Happy to discuss optimizations later once we get the API right though

filter/lua.go
arl: done

filter/lua.go
arl: I'm not sure I see how can the user could be confused here since the config h as 2 fields: a filename and a function name.

filter/lua.go
arl: Since the lua function is in a file, the user could have written some statements before the function. This statement may be needed for the filter to work for example. The current implementation allows that by creating the state once and locking it each time any goroutine enters the lua filter. Recreating the state each time, apart from the potential cost, removes this possibility, since the lua file would need to be reinterpreted each time, that means for each record, at least if I correctly understood what you meant.

Another possible API I thought of was to ask the user to code the lua function directly in the TOML, but you lose syntax highlighting when coding in lua, you'll need to escape your code, etc.

arl: Trying to refocus the discussion a bit.

The idea behind the lua filter is to lower the bar in using baker and coding some filters fast. If I have to code pure lua functions, that means I won't be able to do simple stuff one could expect to be able to do in a filter, such as counting records for example.

Apart from being faster (which, as always with performance, would needs to be proved with numbers) I'm not sure I see any other benefit in requiring the user to write pure functions.

lucarin91: I think that we have to clarify, before going into the implementation details, how we want to execute the LUA code of the user. I see two possible alternatives:

1. sequentially execute the LUA code on the records 
2. parallelize the LUA code on multiple records

The two alternatives have pros and cons and both could drastically change both the performance and the expressiveness of the user. We could also think of a configurable behavior of the filter, but in both cases, the user must be aware of what he/she can and cannot do.

arl: I agree in any case the user must know the tradeoffs that have been made so they can chose if they match their use case.
Coming back to @tommyblue comments about the position of unlock, it seems right to release the lock before calling `next` to allow concurrent execution of the rest of the filter chain. Will fix this and perform the benchmark/tests with this :+1:

Oppen: Any way to offload the locking to the user inside the Lua code? This way they can indicate if they need synchronization or not.

PR #111, opened 2020-12-12
Replace FieldName function to Fieldname slice

arl:
#### :question: What

This replaces `FieldName func(baker.FieldIndex) string` with a simple slice called `FieldNames []string`

This is a tiny breaking change but it brings the following benefits:
 - code is simpler to read and write (see modified examples),
 - though not measured because it's not so important, but this is probably more performant
 - overall it gives the possibility to know the maximum number of defined fields, by just doing len(FieldNames), which was impossible to do before, and needed some times.
 - before, if the user provided an wrong index, the only way to return an error was to return an empty string, which has to be checked anyway. Now the user can avoid that by checking the index against the slice length, and if a wrong index is provided, go will panic with an out of bounds index message, which is way less dangerous that returning an empty string (and thus what could be considered as a valid field name)

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


tommyblue:
you should also update the README and the docs

tommyblue (commented):
please also add something in the changelog as it's a breaking change

arl:
@tommyblue Updated CHANGELOG.md 

Also I opened #114 which contains the website change (to be merged after current PR)

tommyblue:
@arl last thing, update the Readme file as it mentions the FieldName function

tommyblue (approved):
after you've updated the readme file, the pr lgtm

PR #113, opened 2020-12-23
Windows build

guerinoni:
#### :question: What

This PR is for support build under win arch. We can close #113

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,5 +1,5 @@
 name: Go
-on: [push]
+on: [push, pull_request]
 jobs:
   build:
     name: Build
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -22,7 +22,6 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
-
       - name: gofmt
         run: make gofmt
 
@@ -36,7 +35,6 @@ jobs:
         run: |
             go get golang.org/x/tools/cmd/goimports
             test -z "$(goimports -d . | tee /dev/stderr)"
-
       - name: Test
         run: make test
 
@@ -46,7 +44,6 @@ jobs:
         run: |
             make -- --ci-test-cover
             curl -s https://codecov.io/bash | bash
-
       - name: Compile tests/benchmarks
         run: make compile-tests
 
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -1,0 +1,18 @@
+name: Go
+on: [push]
+jobs:
+  build:
+    name: Build
+    runs-on: windows-latest
+    steps:
+      - name: Set up Go 1.15.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15.x
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: build
+        run: go build
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -1,5 +1,5 @@
 name: Go
-on: [push]
+on: [push, pull_request]
 jobs:
   build:
     name: Build
--- a/help_markdown.go
+++ b/help_markdown.go
@@ -4,10 +4,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
-	"runtime"
 	"strings"
-	"syscall"
-	"unsafe"
 
 	"github.com/charmbracelet/glamour"
 )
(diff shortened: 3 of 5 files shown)


tommyblue:
@guerinoni thanks for this PR. I think `runs-on` doesn't like the array of operating systems. You should probably add a job for each os

guerinoni:
@tommyblue now should be fine but I can't see the workflows result :O

tommyblue:
@guerinoni nope, still not working: https://github.com/guerinoni/baker/actions/runs/442485633

.github/workflows/go.yml
```diff
@@ -3,7 +3,11 @@ on: [push]
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04, ubuntu-latest, macos-latest, windows-latest]
```
tommyblue, line 8: either add a mac-specific file or remove from here. And for ubuntu I think we can leave just `ubuntu-latest`

baker_syscall.go
```diff
@@ -0,0 +1,11 @@
+// +build linux darwin
```
arl, line 1: Shouldn't this be `!windows` instead in order to not exclude all oses where baker would currently compile?

arl (changes requested):
Thanks for doing that 
However I'm not sure we should do it in this way.
Ideally, when we want to specialize some piece of code with respect to the os/arch, we should create 2 functions, one for the general case and one for the specialized case (i.e windows here).

For example, a Go function called `termSize()` in a go file (guarded by `!windows`) and another one in a `xxx_windows.go` file.

Ideally, general baker code shouldn't need to be calling `syscall` functions, and syscall functions would only be seen in os/arch guarded files.

baker_syscall.go
guerinoni: It seems better!

arl:
@guerinoni you only addressed a part of the comments I made. If my comment about the `termSize` function is not clear I can expand a bit or we could have a fast call about that.

arl:
@guerinoni can you expand a bit as to why a darwin github workflow file was needed? In what is it different from, say, the linux one?
And in the same manner, why a windows-specific one is required?

guerinoni:
@arl the platforms are different and it can be useful building application under all 3 main platforms, in future something can be different

arl:
> @arl the platforms are different and it can be useful building application under all 3 main platforms, in future something can be different

I agree with you but I'd prefer the PR to stay focused on making baker build on windows, so I don't see why darwin is involved. Also it could be nice, if we add a windows-specific workflow, to have the ci run baker tests and not just build.
Once this is done, we can discuss in another pr about adding darwin-specific ci build.

Also, once the lua filter PR gets merged we can also discuss about making girhub-ci creates artifacts, baker binaries one can download and use directly.

arl (approved):
Thanks @guerinoni 
LGTM!

.github/workflows/win.yml
```diff
@@ -0,0 +1,18 @@
+name: Go
+on: [push]
```
arl, line 2: @guerinoni maybe the windows build is not starting because the branch is coming from a pull-request (from a fork). Your PR is the first one coming from a fork as before we were only creating pull-requests from branches in the current repository (adroll).
Do you think that we should add 
```
on: [push, pull_request]
```
to make it work?
If that 's true then we should add it to all github actions workflows

guerinoni: From docs
```
# Triggered when code is pushed to any branch in a repository
on: push
```

.github/workflows/go.yml
```diff
@@ -46,12 +44,11 @@ jobs:
...
-
       - name: Compile tests/benchmarks
         run: make compile-tests
 
       - name: Build all examples
         run: go build ./examples/...
 
       - name: Build with extras
-        run: CGO_ENABLED=1 go build -tags cgo_sqlite ./examples/advanced/
+        run: CGO_ENABLED=1 go build -tags cgo_sqlite ./examples/advanced/
```
tommyblue, line 54: all changes to this file can be removed

.github/workflows/win.yml
arl: Indeed but the source  branch of your PR comes from your forked repository.
I'm not sure about this but I think we should try since at the moment no github actions are built, whereas they always did before when PR branches were located on AdRoll repository;

arl: After reading the docs more thoroughly it indeed seems that we're missing `pull_request` if we want workflows to be triggered from PR coming from forked repositories.
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

arl:
Thank you!

PR #118, opened 2021-01-26
fix: list input leave readers opened

lucarin91:
#### :question: What

In the `List` input some io Readers were left open. 
- Function `openFile` was returning the io Reader also when the `onlySize` parameter was true, which create a leak and leave open the io Reader of the file.
- Function `processListFile` does not close the file after reading it.


input/list.go
```diff
@@ -215,6 +224,7 @@ func (s *List) processListFile(f io.ReadCloser) {
 
 	go func() {
 		scanner := bufio.NewScanner(f)
+		defer f.Close()
```
arl, line 227: I seem to have missed this the first time.

`processListFile` is used 4 times in the code base:
 - once in case `s3` and `f` is closed by the caller.
 - once in case `http` and `f` is closed by the caller.
 - once in case `-` (for stdin) `f` isn't closed by the caller (and it's ok).
 - once in case `file` and `f` isn't closed by the caller (which is a bug).

I admit the API is ambiguous: since `f` is a `ReadCloser`, one expects to have to close it, however as it's the case with stdin we don't always want to do it.

I think that the best way to solve this is to change the `processListFile` prototype to make it accept an `io.Reader`, not even giving the possibliity for us to close so it becomes obvious that it's the caller responsibility to do it if needed.

arl (changes requested):
See my comment about `processListFile`

arl (approved):
Ok, tested :+1:

PR #122, opened 2021-02-02
feat: record validation in TOML

lucarin91:
#### :question: What

Make it possible to declare Record validation from TOML. A new `Validation` key is added in the TOML, where the user can set a regex for each of the field names.

```
[Validation]
name = “^[a-z]*$“
timestamp = “^[0-9]*“
```

To avoid conflicts the validation could be omitted or set just one time either on the Component object or in the TOML.

#### :hammer: How to test

1. Add the Validation key on a Baker TOML
2. Make sure that no validation function is passed during the configuration
3. Records should be correctly validated and failed one throw away


e2e_test.go
```diff
@@ -25,6 +25,12 @@ func TestE2EFullTopology(t *testing.T) {
 		"./examples/basic/", "", "", "",
 	))
 
+	t.Run("validation", testE2EFullTopology(
+		"./examples/validation/", "",
+		"_out/output.csv.gz",
+		"testdata/output.csv.gz.golden",
+	))
+
```
arl, line 33: This example doesn't really illustrate nor test the new validation feature. Both files (input and output) contain:

```csv
value1,value2,value3
value1,value2,value3
value5,value2,value3
```
so no record has been invalidated by the validation function.

Here we could for example use an already existing input file, say `testdata/input.csv.zst`, but make it so that not all records pass the validation function generated from `validation/main.go`.
At this point the output file will be different, so you can add new `golden` file, `./testdata/output.validation.csv.gz.golden` for example, into the repository, and update the end-to-end with that new file

examples/validation/main.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func main() {
+	toml := `
+[fields]
+	names=["timestamp", "source", "target"]
+
+[validation]
+	timestamp="^value[0-9]+$"
+	target="value3"
```
arl, line 22: As explained above (in e2e_test.go comment) we should generate a validation function which makes at least one record _not pass_ the check

arl (changes requested):
Some comments here and there to update and the end-to-end test also, but the non-test code looks good to me.

Good job :+1:

e2e_test.go
lucarin91: The input file of the example, i.e. `testdata/input.csv.zst`, contains:
```
value1,value2,value3
value1,value2,value3
value4,value2,value6
value5,value2,value3
```
Therefore the validation remove the record `(value4,value2,value6)`. I think that the example correctly illustrates the feature: the `timestamp` could be any `value+a number`, `source` is not validated and `target` should contain `value3`.
Anyway, if you think that the example should be more clear I am open to suggestions.

Moreover, I take your suggestion and I will modify the output file from `output.csv.gz.golden` in `output.validation.csv.gz.golden`.

arl: my bad I opened the wrong file. :+1:

arl (approved):
LGTM

PR #123, opened 2021-02-03
fix: assignFieldMapping test skip a case

lucarin91:
#### :question: What

The test case `duplicated field name`  of `Test_assignFieldMapping` does not correctly check the intended case because a Components is passed. The function will be stopped with the error `field indexes/names can't both be set in TOML and Components` and not `duplicated field name "baz"`.

Here the coverage screenshot:

Screenshot from 2021-02-03 10-21-44


arl (approved):
Thanks, good catch!

PR #128, opened 2021-02-10
feat: add ExpandJSON filter

lucarin91:
#### :question: What

Add the `ExpandJSON` filter. The filter copy the values of a set JSON keys to corresponding record fields.
The filter support also [JMESpath](https://jmespath.org/tutorial.html) to select the values to copy inside the JSON.

A possible filter configuration is:

```
[[filter]]
name="ExpandJSON"
	[filter.config]
	Source = "json_data"

	[filter.config.Fields]
	jfield1  = "field1"
        jfield2  = "field2"
```

The filter will transform the following input in the corresponding output:

__Input:__

| field1 | field2 | json_data                                     |
|:-------:|:-------:|:-------------------------------------------:|
|          |           | {jfield1:"value1", jfield2:"value2"} |

__Output:__

| field1   | field2  | json_data                                     |
|:---------:|:--------:|:-------------------------------------------:|
| value1 | value2 | {jfield1:"value1", jfield2:"value2"} |

#### :hammer: How to test

1. Configure the filter in the TOML
2. Change the CSV delimiter to something like `;` to avoid a clash with the JSON
3. Run Baker with an appropriate input


lucarin91:
### UPDATE: add JMESpath support

The last version of the PR uses the library [`go-jmespath`](https://github.com/jmespath/go-jmespath) for selecting the JSON value to copy.
This solution will increase both the versatility of the filter and the performance.

Here the performance comparisons between the naive implementation with a map and the actual implementation with JMESpath:

```
name          old time/op    new time/op    delta
ExpandJSON-8    3.01µs ±10%    1.70µs ±13%  -43.52%  (p=0.000 n=10+10)

name          old alloc/op   new alloc/op   delta
ExpandJSON-8    1.97kB ± 0%    1.38kB ± 0%  -29.67%  (p=0.000 n=10+10)

name          old allocs/op  new allocs/op  delta
ExpandJSON-8      31.0 ± 0%      20.0 ± 0%  -35.48%  (p=0.000 n=10+10)
```

filter/expand_json.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/AdRoll/baker"
+	"github.com/jmespath/go-jmespath"
+)
+
+var ExpandJSONDesc = baker.FilterDesc{
+	Name:   "ExpandJSON",
+	New:    NewExpandJSON,
+	Config: &ExpandJSONConfig{},
+	Help:   "Explodes json objects to other fields.",
```
arl, line 17: - We can expand a bit the help string, maybe also including a link to https://jmespath.org/.
you can use multi-lines here if needed.

- Not sure about using explodes here

filter/expand_json.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type ExpandJSONConfig struct {
+	Fields          map[string]string `help:"<JMESPath -> record field> map, the rest will be ignored"`
+	Source          string            `help:"record field that contains the json"`
```
arl, line 22: This field is required for the filter to function properly. You inform baker (and also the user when they prints the filter help) by adding the required struct tag, like so:
```suggestion
	Source          string            `help:"record field that contains the json" required:"true"`
```

filter/expand_json.go
```diff
@@ -0,0 +1,133 @@
...
+		ut.jexp = append(ut.jexp, c)
+	}
+	// TrueFalseValues
+	if l := len(dcfg.TrueFalseValues); l != 2 {
+		return nil, fmt.Errorf("only two True False values allowed, %v given", l)
+	}
+	ut.trueFalseValues = [2][]byte{
+		[]byte(dcfg.TrueFalseValues[trueIdx]),
+		[]byte(dcfg.TrueFalseValues[falseIdx]),
+	}
```
arl, line 78: form: this whole block of lines of code seems to be screaming for a bit of air (a.k.a some extra line here and there to group the code in logical blocks won't hurt the reader's eyes) :-)

filter/expand_json.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := d.Decode(&x); err != nil {
+		return nil
+	}
+	return x
+}
+
+func (f *ExpandJSON) postProcessJSON(r interface{}) (val []byte) {
+	switch typedValue := r.(type) {
+	case json.Number:
+		val = []byte(typedValue)
```
arl, line 120: using a named return value in this method doesn't seem really needed, also the code would probably be easier to read if we were just returning instead of setting `val = `

filter/expand_json_test.go
```diff
@@ -0,0 +1,195 @@
...
+					i, _ := fieldByName(k)
+					if !bytes.Equal(rec2.Get(i), []byte(v)) {
+						t.Errorf("got %q, want %q", rec2.Get(i), v)
+					}
+				}
+				// check that the json field is untouched
+				i, _ := fieldByName("j")
+				if !bytes.Equal(rec2.Get(i), []byte(tt.json)) {
+					t.Errorf("got %q, want %q", rec2.Get(i), tt.json)
+				}
```
arl, line 149: this is probably overly defensive. I agree record.Get should not modify the record, but it's not that test job to verify that but a test of the specific Record implementation

filter/expand_json_test.go
```diff
@@ -0,0 +1,195 @@
...
+		{
+			name:      "test boolean 2",
+			json:      `{"j1": true, "j2": false}`,
+			want:      map[string]string{"f1": "t", "f2": "f"},
+			trueFalse: []string{"t", "f"},
+		},
+		{
+			name: "other json types",
+			json: `{"j1": null, "j2": [1,2,3,4], "j3": {"j1":"a", "j2":false}}`,
+			want: map[string]string{"f1": "", "f2": "[1,2,3,4]", "f3": `{"j1":"a","j2":false}`},
```
arl, line 49: nit: wouldn't it be a bit more readable if this was like:
```suggestion
			want: map[string]string{
				"f1": "",
				"f2": "[1,2,3,4]",
				"f3": `{"j1":"a","j2":false}`,
			},
```

filter/expand_json_test.go
```diff
@@ -0,0 +1,195 @@
...
+			if len(tt.field) == 0 {
+				tt.field = map[string]string{
+					"j1": "f1",
+					"j2": "f2",
+					"j3": "f3",
+				}
+			}
+			if len(tt.source) == 0 {
+				tt.source = "j"
+			}
```
arl, line 111: why is this needed?

arl (approved):
Most of my comments are about the form, code is correct.

Good speedup of this filter with much more features.
Good job :+1:

filter/expand_json_test.go
lucarin91: It is needed to set the default value of for the source. I will change it in `tt.source == ""` maybe it will be more clear.

PR #138, opened 2021-02-25
feat: support map types in PrintHelp

lucarin91:
#### :question: What

Add support of table/map in the automatically generated documentation.
Supported types:
- table of string `map[string]string`
- table of ints `map[string]int`

The PR also add a new test, which checks that `PrintHelp` does not return an error during the documentation generation.


help_config.go
```diff
@@ -177,8 +177,22 @@ func newHelpConfigKeyFromField(f reflect.StructField) (helpConfigKey, error) {
...
+			switch f.Type.Elem().Kind() {
+			case reflect.String:
+				h.typ = "table of strings"
+			case reflect.Int:
+				h.typ = "table of ints"
+			default:
+				return h, fmt.Errorf("config key %q: unsupported type table of %s", f.Name, f.Type.Elem())
+			}
+		default:
+			return h, fmt.Errorf("config key %q: unsupported map with key type %s", f.Name, f.Type.Key())
```
arl, line 192: I know the TOML standard calls these _tables_ but I think we should stick to `map` in the help message

help_test.go
```diff
@@ -54,3 +56,16 @@ func TestAllUploadsHaveConfigHelp(t *testing.T) {
...
 }
+
+func TestPrintHelp(t *testing.T) {
+	comp := baker.Components{
+		Inputs:  input.All,
+		Filters: filter.All,
+		Outputs: output.All,
+		Uploads: upload.All,
+	}
+	err := baker.PrintHelp(&strings.Builder{}, "*", comp, baker.HelpFormatMarkdown)
```
arl, line 67: you can use `ioutil.Discard` here instead of a `string.Builder`.

arl (approved):
LGTM!

help_config.go
lucarin91: I was opting for the table naming because a table in the TOML world implicitly uses strings as keys. If we say `map of strings` and `map of ints` we may not be sufficiently precise about the type. Maybe we should use `map of strings to strings` and `map of strings to ints`. I am open to suggestions.

arl: `map of strings to strings` and `map of strings to ints` is good

arl (approved):
:+1:

PR #134, opened 2021-02-24
doc: ExpandJSON filter

lucarin91:
#### :question: What

Update site for PR #128 

#### :white_check_mark: Checklists

- [x] Have the steps in [CONTRIBUTING.md](./CONTRIBUTING.md) been followed to update a Go module?

PR #141, opened 2021-02-25
doc: improve ExpandJson filter doc

lucarin91:
#### :question: What

Improve `ExpandJson` filter documentation with @tommyblue suggestions in #134.


filter/expand_json.go
```diff
@@ -10,11 +10,48 @@ import (
...
+		jfield1  = "field1"
+		jfield2  = "field2"
+
+The filter will transform the following input in the corresponding output:
+
+**Input:**
+
+| field1 | field2 |              json_data                 |
+| :----: | :----: | :------------------------------------: |
+|        |        | ` + "`{\"jfield1\":\"value1\", \"jfield2\":\"value2\"}`" + ` |
```
arl, line 41: It's a good idea to give an example for a filter which is somewhat more involved.
Maybe using a markdown table here is a bit overkill?
Just showing the TOML, what contains `json_data` and explaining that the filter outcome is to write `value1` into `field1` and `value2` into `field2` might be enough.

arl: Also, we don't need backslashes to represent a valid json string. Also I think we could drop the double quotes around `json_data` content and just write `| {"jfield1":"value1", "jfield2":"value2"} |`

filter/expand_json.go
```diff
@@ -10,11 +10,48 @@ import (
 	"github.com/jmespath/go-jmespath"
 )
 
+const help = `
```
arl, line 13: rename this to `expandJSONhelp` to not _steal_ the `help` variable name to all code located in the `filter` package

arl (changes requested):

filter/expand_json.go
```diff
@@ -10,11 +10,32 @@ import (
...
+
+	[[filter]]
+	name="ExpandJSON"
+		[filter.config]
+		Source = "json_data"
+		[filter.config.Fields]
+		jfield1  = "field1"
+		jfield2  = "field2"
+		
+In this example, the filter will extract the value of the key ` + "`jfield1`" + ` and ` + "`jfield2`" + ` of the JSON
```
arl, line 29: Generally using present tense in technical writings makes for _lighter_ and easier to read sentences. 
```suggestion
In this example, the filter extracts values of the ` + "`jfield1`" + ` and ` + "`jfield2`" + ` keys of the JSON 
```

filter/expand_json.go
```diff
@@ -10,11 +10,32 @@ import (
...
+	[[filter]]
+	name="ExpandJSON"
+		[filter.config]
+		Source = "json_data"
+		[filter.config.Fields]
+		jfield1  = "field1"
+		jfield2  = "field2"
+		
+In this example, the filter will extract the value of the key ` + "`jfield1`" + ` and ` + "`jfield2`" + ` of the JSON 
+object present in field ` + "`json_data`" + `of the record. Then, the valus of that keys will be written into the field
```
arl, line 30: typo:
```suggestion
object present in field ` + "`json_data`" + `of the record. Then, the values of that keys will be written into the field 
```

arl (approved):
Last 2 comments, I promise :angel: 
Once they're addressed, you can merge directly without another review from me

PR #142, opened 2021-02-26
chore: fix CI cache

lucarin91:
#### :question: What

Configure the CI cache as suggested [here](https://github.com/actions/cache/blob/main/examples.md#go---modules).
The previous configuration restore, all the time, an empty cache.

### DIFF
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -17,7 +17,8 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: |
-            ~/go/pkg/mod              # Module download cache
+            ~\go\pkg\mod
+            %LocalAppData%\go-build
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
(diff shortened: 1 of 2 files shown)


.github/workflows/win.yml
```diff
@@ -17,7 +17,8 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: |
-            ~/go/pkg/mod              # Module download cache
+            ~\go\pkg\mod
+            ~\AppData\Local\go-build
```
arl, line 21: The linux syntax above is ok, but for Windows this is not in line with what I found on https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds:
 - the use of backslashes
 - `%LocalAppData%\go-build`

arl (commented):
Look at the comments for the Windows actions

.github/workflows/win.yml
lucarin91: For what I see in the [cache examples](https://github.com/actions/cache/blob/main/examples.md), `~\AppData\Local` and `%LocalAppData%` are the same. I think that is just a matter of taste. If you prefer I will switch to:
```
  ~\go\pkg\mod
  %LocalAppData%\go-build
```

arl: Yeah better using the most generic one,  %LocalAppData% then

arl (approved):
LGTM!

PR #143, opened 2021-02-26
feat: add Dedup filter

lucarin91:
#### :question: What

Add the `Dedup` filter.  `Dedup` filter out records which contains identical values on configurable fields.


filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"github.com/AdRoll/baker"
+)
+
+const dedupHelp = `
+Removes record with identical value using a set of provided fields as key.
```
arl, line 13: I'd make the word duplicate appear in the first sentence, to justify the filter name immediately. Also I'd avoid explaining too much about implementation details, but focus on usage and limitation, something along the lines of:
```suggestion
This filter removes duplicate records. A record is considered a duplicate, and is thus removed by this filter, 
if another record with the same values has already been _seen_. The comparison is performed on a user-provided list of fields (Fields setting).
```

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+
+The deduplication procedure internally uses a shared set, concurrently used between all the filter replicas for storing the encountered record key.
+Indeed, the use of this filter could introduce performance degradation and possible unbounded memory consumption.
+
+The user is responsible to properly choose the set of fields to construct a bounded combination of possible inputs and thus maintaining the
+memory consumption under control.
```
arl, line 19: Good idea to give a warning about unbound memory consumption.
As explained above, I wouldn't go too much into implementation details here, they might change in the future. However I'd give the user enough information for them to understand what's happening and consider using or not this filter.

For example:
```suggestion
WARNING: to remove duplicates, this filter stores one key per unique record in memory, this means that the overall memory grows linearly with the number of unique records in your data set. Depending on your data set, this might 
lead to OOM (i.e out of memory) errors.
```

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+
+var DedupDesc = baker.FilterDesc{
+	Name:   "Dedup",
+	New:    NewDedup,
+	Config: &DedupConfig{},
+	Help:   dedupHelp,
+}
+
+type DedupConfig struct {
+	Fields []string `help:"fields that needs to be unique" required:"true"`
```
arl, line 30:
```suggestion
	Fields []string `help:"fields to consider when comparing records" required:"true"`
```

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+	_, found := f.dedupSet.LoadOrStore(key, struct{}{})
+
+	if found {
+		atomic.AddInt64(&f.numFilteredLines, 1)
+	} else {
+		next(l)
+	}
+}
+
+// constructKey build a key with the concatenation of the fields
```
arl, line 83:
```suggestion
// constructKey builds a key by concatenating field values
```

arl (approved):
Code LGTM!
Comments/doc could be improved

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	dcfg := cfg.DecodedConfig.(*DedupConfig)
+
+	f := &Dedup{
+		cfg: dcfg,
+	}
+	for _, field := range dcfg.Fields {
+		i, ok := cfg.FieldByName(field)
+		if !ok {
+			return nil, fmt.Errorf("unrecognized deduplication field %q", field)
```
arl, line 56: Generally for this kind of errors, we use this. Keeping a kind of homogeneity between error messages also helps the user:
```suggestion
			return nil, fmt.Errorf("unknown field %q", field)
```

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+
+	key := f.constructKey(l)
+	_, found := f.dedupSet.LoadOrStore(key, struct{}{})
+
+	if found {
+		atomic.AddInt64(&f.numFilteredLines, 1)
+	} else {
+		next(l)
+	}
+}
```
arl, line 81: style: keep the happy path to the right and unindent the happy path, using early returns if needed
```suggestion
	if _, found := f.dedupSet.LoadOrStore(key, struct{}{}); found {
		atomic.AddInt64(&f.numFilteredLines, 1)
		return
	}

	next(l)
}
```

filter/inc_dedup.go
```diff
@@ -0,0 +1,90 @@
...
+
+// constructKey builds a key by concatenating field values
+func (f *Dedup) constructKey(l baker.Record) string {
+	var sb strings.Builder
+	for _, i := range f.fields {
+		sb.Write(l.Get(i))
+	}
+	return sb.String()
+}
```
arl, line 90: In theory you could remove a record which you shouldn't with this kind of pathological case:

 - record 1: `ciao,hello`
 - record 2: `ci,aohello`
 
both will have the same _key_ `ciaohello` if you just concatenate. That's probably why the previous version of the filter used `,` in between the values. We should keep this behavior.
Also I suggest you to write a test before any modification proving that this case fails at the moment, then fix it

arl (changes requested):
See my comment about key generation.
We should also rename the files into `dedup.go` and `dedup_test.go`

filter/inc_dedup.go
lucarin91: Yes, you are right this is a very critical mistake, I am sorry for that.

filter/dedup.go
```diff
@@ -0,0 +1,105 @@
...
+		i, ok := cfg.FieldByName(field)
+		if !ok {
+			return nil, fmt.Errorf("unknown field %q", field)
+		}
+		f.fields = append(f.fields, i)
+	}
+
+	sep := []rune(dcfg.KeySeparetor)
+	if len(sep) != 1 || sep[0] > unicode.MaxASCII {
+		return nil, fmt.Errorf("separetor must be a 1-byte string or hex char")
```
arl, line 72:
```suggestion
		return nil, fmt.Errorf("separator must be a 1-byte string or hex char")
```

filter/dedup.go
```diff
@@ -0,0 +1,105 @@
...
+var DedupDesc = baker.FilterDesc{
+	Name:   "Dedup",
+	New:    NewDedup,
+	Config: &DedupConfig{},
+	Help:   dedupHelp,
+}
+
+type DedupConfig struct {
+	Fields       []string `help:"fields to consider when comparing records" required:"true"`
+	KeySeparetor string   `help:"character separator used to build a key from the fields" default:"\x1e"`
```
arl, line 32:
```suggestion
	KeySeparator string   `help:"character separator used to build a key from the fields" default:"\x1e"`
```

filter/dedup_test.go
```diff
@@ -0,0 +1,156 @@
...
+		{
+			name: "not existing field",
+			records: []string{
+				"abc,def,ghi",
+			},
+			fields:  []string{"not_exist"},
+			wantErr: true,
+		},
+		{
+			name: "separetor more 1-byte",
```
arl, line 89:
```suggestion
			name: "separator more 1-byte",
```

filter/dedup_test.go
```diff
@@ -0,0 +1,156 @@
...
+			records: []string{
+				"abc,def,ghi",
+			},
+			fields:  []string{"f1"},
+			sep:     "ab",
+			wantErr: true,
+		},
+		{
+			name: "separetor over max ASCII",
```
arl, line 98:
```suggestion
			name: "separator over max ASCII",
```

arl (approved):
LGTM!

apart from separetor -> separator :-)
you can merge once fixed

PR #148, opened 2021-03-03
refactor: nil check in component constructor

lucarin91:
#### :question: What

In the Baker components instantiation function, there is always a check if `cfg.DecodedConfig` is not nil. Although the test will always be false:
- the config structure is created in the description of each component, e.g., [ ClauseFilterDesc](https://github.com/AdRoll/baker/blob/main/filter/clausefilter.go#L75), and
- the config structure is set to `cfg.DecodedConfig` in `config.go`, i.e., [here](https://github.com/AdRoll/baker/blob/main/config.go#L314) and [here](https://github.com/AdRoll/baker/blob/main/config.go#L361)
 
This is true even if the user does not specify any filter configuration in the TOML. I think that removing these checks will make the code of the components more clear.


arl (commented):
That's a great improvement both in terms of readability and writing new components.
However, I'd like to see some unit-tests (in the baker package) written that check that:
 - one can create a `xxxDesc` struct variable in which the `Config` field is a `&struct{}{}` (for components which don't require any configuration)
 - check that this works for `InputDesc`, `OutputDesc`, `FilterDesc` as well as `UploadDesc`
 - it would be even better if the same can also be applied to `UserDesc` and `MetricsDesc`

desc_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Fatalf("cannot build topology: %v", err)
+	}
+}
+
+type dummyInput struct{}
+
+func (dummyInput) Run(chan<- *baker.Data) error { return nil }
+func (dummyInput) Stop()                        {}
+func (dummyInput) Stats() baker.InputStats      { return baker.InputStats{} }
+func (dummyInput) FreeMem(*baker.Data)          {}
```
arl, line 109: Instead of declaring this you can directly use `inputtest.Base` see https://pkg.go.dev/github.com/AdRoll/baker@v0.0.0-20210303153518-3ff24dd567ba/input/inputtest#Base
Same for output, filter and upload
For the metrics, there's https://pkg.go.dev/github.com/AdRoll/baker#NopMetrics

arl (approved):
LGTM!

You can merge this once you have simplified `TestCompDescEmptyConfig` in `desc_test.go`

PR #151, opened 2021-03-03
feat: add FormatTime filter

lucarin91:
#### :question: What

Add the `FormatTime` filter.


filter/format_time.go
```diff
@@ -0,0 +1,194 @@
...
+			return strconv.AppendInt(nil, t.UnixNano(), 10)
+		}
+	default:
+		layout := formatToLayout(dcfg.DstFormat)
+		f.format = func(t time.Time) []byte {
+			return []byte(t.Format(layout))
+		}
+	}
+
+	return f, nil
```
arl, line 146: Before returning the instantiated filter, we should ensure that the user-provided formats are valid. I'm not sure all user-provided layout strings are valid ones so, since it's something we can know now, better to return an error and make baker fail early, rather than once the topology has started fetching records, etc, etc.

You can do that by simply using time.Now() with the `f.parse` and `f.format` closures. Also this should be present in unit-tests

lucarin91: I already tried to find a way to check the time format of the user, but I think that it is not possible, see [playground](https://play.golang.org/p/MG7znIVcXLJ). The `time.Parse` function fails only if the layout and the time string do not match. The problem is that we cannot know the inputs of the user in advance.

filter/format_time.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			return time.Unix(sec, 0), nil
+		}
+	case unixms:
+		return func(b []byte) (time.Time, error) {
+			msec, err := strconv.ParseInt(string(b), 10, 64)
+			if err != nil {
+				return time.Time{}, err
+			}
+			return time.Unix(0, msec*1_000_000), nil
```
arl, line 176: there are constants in the `time` package for that:
```suggestion
			return time.Unix(0, msec*int64(time.Millisecond)), nil
```

filter/format_time_test.go
```diff
@@ -0,0 +1,294 @@
...
+			fin:  "unix",
+			out:  refTime.Format(lout),
+			fout: fout,
+		}
+		tests = append(tests, t)
+	}
+	for fout, lout := range format {
+		t := test{
+			name: "unixms->" + fout,
+			in:   fmt.Sprintf("%d", refTime.UnixNano()/1_000_000),
```
arl, line 66:
```suggestion
			in:   fmt.Sprintf("%d", refTime.UnixNano()/int64(time.Millisecond)),
```

filter/format_time_test.go
```diff
@@ -0,0 +1,294 @@
...
+			fout: "unix",
+		}
+		tests = append(tests, t)
+	}
+	for fin, lin := range format {
+		t := test{
+			name: fin + "->unixms",
+			in:   refTime.Format(lin),
+			fin:  fin,
+			out:  fmt.Sprintf("%d", timeAsFrom(lin, refTime).UnixNano()/1_000_000),
```
arl, line 99:
```suggestion
			out:  fmt.Sprintf("%d", timeAsFrom(lin, refTime).UnixNano()/int64(time.Millisecond)),
```

arl (approved):
Good job :+1: 
nice filter!

you can merge this after addressing the last comments

PR #154, opened 2021-03-04
Fix data race in statsDumper

arl:
#### :question: What

There was a data race in stats dumper (found by running baker compiled with -race). The data race could potentially corrupt the metric about the number of invalid fields, which is a minor metric 

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


stats.go
```diff
@@ -129,6 +129,7 @@ func (sd *StatsDumper) dumpNow() {
 
 	if invalid > 0 {
 		m := make(map[string]int64)
+		t.mu.Lock()
```
lucarin91, line 132: Is not sufficient the `RLock` here?

arl: This lock is taken once a second so it would probably not change anything to complicate the semantics here.
I don't even know why this is a rwmutex, it should be a simple mutex since the write lock (in topology.go) is the one we could potentially do more often. And if we don't have any invalid fields, then the lock is only ever acquired once per second, which is nothing

lucarin91: Yes, you are right. But I saw that [here](https://github.com/AdRoll/baker/blob/6f870df8405b6f4fbefa5136f357d03773c102e3/stats.go#L171) in the same file the `RLock()` is used to do the same thing. Maybe we could unify the use.

arl: yeah there's probably a way to loop only once over the map. But at this point it's more of a cosmetics change since the use is not problematic, anyway it's not the target of this pull-request. I create a ticket on our jira board for the unification

tommyblue (approved):

PR #155, opened 2021-03-04
feat: add ExpandList filter

lucarin91:
#### :question: What

Add ExpandList filter


filter/expand_list.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (f *ExpandList) Process(l baker.Record, next func(baker.Record)) {
+	atomic.AddInt64(&f.numProcessedLines, 1)
+
+	list := l.Get(f.source)
+	part := bytes.Split(list, f.sep)
+	for i, idx := range f.listIdx {
+		if idx >= len(part) {
+			continue
+		}
```
tommyblue, line 116: you could create listIdx as a sorted list (and field as a map fieldIdx->field) and then break here instead of continuing. it would be heavier while starting but faster while looping over the records. what do you think?

lucarin91: Yes, good idea. But it is not necessary to have a map `fieldIdx->fields`, we just need to sort both slices in the same way.

tommyblue: :+1:

tommyblue (approved):

PR #162, opened 2021-03-09
doc: update MetricsBag

lucarin91:
#### :question: What

Update the documentation of the `MetricsBag`.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -19,7 +19,7 @@ func (bag MetricsBag) AddRawCounter(name string, value int64) {
 
 // AddDeltaCounter adds a new delta increment for a counter. The increment represents
 // the variation of the counter in the current time interval.
-// delta must be positive.
+// To be meaningful, `delta` must be positive.
 func (bag MetricsBag) AddDeltaCounter(name string, delta int64) {
 	bag["d:"+name] = delta
 }
--- a/metrics.go
+++ b/metrics.go
@@ -13,14 +13,14 @@ type MetricsBag map[string]interface{}
 
 // AddRawCounter adds a metric of type counter to the bag and sets its current value.
 // A counter is a cumulative metric that can only increase.
-// To be meaningful, `value` must be positive.
+// To be meaningful, value must be positive.
 func (bag MetricsBag) AddRawCounter(name string, value int64) {
 	bag["c:"+name] = value
 }
 
 // AddDeltaCounter adds a metric of type counter to the bag and increments its value.
 // A counter is a cumulative metric that can only increase.
-// To be meaningful, `delta` must be positive.
+// To be meaningful, delta must be positive.
 func (bag MetricsBag) AddDeltaCounter(name string, delta int64) {
 	bag["d:"+name] = delta
 }
--- a/metrics.go
+++ b/metrics.go
@@ -11,8 +11,8 @@ import (
 // with values relative to that last second.
 type MetricsBag map[string]interface{}
 
-// AddRawCounter adds a counter. A counter is a cumulative metrics that can only increase.
-// value must be positive.
+// AddRawCounter adds a counter. A counter is a cumulative metric that can only increase.
+// To be meaningful, `value` must be positive.
 func (bag MetricsBag) AddRawCounter(name string, value int64) {
 	bag["c:"+name] = value
 }
--- a/metrics.go
+++ b/metrics.go
@@ -11,40 +11,42 @@ import (
 // with values relative to that last second.
 type MetricsBag map[string]interface{}
 
-// AddRawCounter adds a counter. A counter is a cumulative metric that can only increase.
+// AddRawCounter adds a metric of type counter to the bag and sets its current value.
+// A counter is a cumulative metric that can only increase.
 // To be meaningful, `value` must be positive.
 func (bag MetricsBag) AddRawCounter(name string, value int64) {
 	bag["c:"+name] = value
 }
 
-// AddDeltaCounter adds a new delta increment for a counter. The increment represents
-// the variation of the counter in the current time interval.
+// AddDeltaCounter adds a metric of type counter to the bag and increments its value.
+// A counter is a cumulative metric that can only increase.
 // To be meaningful, `delta` must be positive.
 func (bag MetricsBag) AddDeltaCounter(name string, delta int64) {
 	bag["d:"+name] = delta
 }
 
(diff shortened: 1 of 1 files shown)


metrics.go
```diff
@@ -11,27 +11,35 @@ import (
 // with values relative to that last second.
 type MetricsBag map[string]interface{}
 
-// AddRawCounter adds a counter that always increments.
+// AddRawCounter adds a counter. A counter is a cumulative metrics that can only increase.
```
tommyblue, line 14:
```suggestion
// AddRawCounter adds a counter. A counter is a cumulative metric that can only increase.
```

metrics.go
```diff
@@ -11,27 +11,35 @@ import (
...
+// value must be positive.
```
tommyblue, line 15: it seems we enforce a check, but we don't, so maybe I'd write:

```suggestion
// To be meaningful, `value` must be positive.
```

metrics.go
```diff
@@ -11,27 +11,35 @@ import (
...
 func (bag MetricsBag) AddRawCounter(name string, value int64) {
 	bag["c:"+name] = value
 }
 
-// AddDeltaCounter adds a count of something that happened in the last second
+// AddDeltaCounter adds a new delta increment for a counter. The increment represents
+// the variation of the counter in the current time interval.
+// delta must be positive.
```
tommyblue, line 22: same as above

metrics.go
```diff
@@ -11,27 +11,35 @@ import (
...
 
-// AddGauge takes a snapshot of a value.
+// AddGauge adds a new snapshot of a value. A Gauge represents a single numerical
+// data point that can arbitrarily go up and down.
 func (bag MetricsBag) AddGauge(name string, value float64) {
 	bag["g:"+name] = value
 }
 
 // AddHistogram adds a set of values to track their statistical distribution.
+// A histogram samples observations and counts them in different 'buckets' in order
```
tommyblue, line 34:
```suggestion
// A histogram sample observes and counts them in different 'buckets' in order
```
Not sure I fully understand what you mean and thus if my suggestion is correct

metrics.go
```diff
@@ -11,27 +11,35 @@ import (
...
+// to track and show the statistical distribution of a set of values.
 func (bag MetricsBag) AddHistogram(name string, values []float64) {
 	bag["h:"+name] = values
 }
 
 // AddTimings adds a set of timings to track their statistical distribution.
+// A histogram samples observations and counts them in different 'buckets'. Timings
```
tommyblue, line 41: same as above

tommyblue (approved):
lgtm, check the language suggestions

metrics.go
lucarin91: `sample` is the verb and `observation` the noun. Thus, I think that was correct.

arl (changes requested):
I'd expect the doc strings for `AddRawCounter` and `AddDeltaCounter` to be very similar. They both act on the same underlying metric of type **counter** so I'd expect both to hold the definition of a counter, while making it easy to understand their differences.

> A counter is a cumulative metric that can only increase.

Example:
```
AddRawCounter adds a new metric of type counter to the bag and sets its current value.
A counter is a cumulative metric that can only increase. 
```

```
AddDeltaCounter adds a new metric of type counter to the bag and increments its value.
A counter is a cumulative metric that can only increase. 
```

Also I'd add to the other metrics _to the bag_ or _into the bag_. That may seems obvious when you know it but I'd prefer this to be explicit, so it shows that a bag is just a container for metrics, it just stores them but does not publish them in any way.

Last thing, we should improve the documentation of the `Merge` method, explaining how Merge aggregates metrics with the same name coming from 2 different bags.

metrics.go
```diff
@@ -11,32 +11,42 @@ import (
 // with values relative to that last second.
 type MetricsBag map[string]interface{}
 
-// AddRawCounter adds a counter that always increments.
+// AddRawCounter adds a metric of type counter to the bag and sets its current value.
+// A counter is a cumulative metric that can only increase.
+// To be meaningful, `value` must be positive.
```
arl, line 16: In general we don't use backticks in comments as anyway they're not rendered in GoDoc.
So both of the following suggestions are ok (I personally prefer the first one):
```suggestion
// To be meaningful, value must be positive.
```
```suggestion
// To be meaningful, 'value' must be positive.
```

metrics.go
```diff
@@ -11,32 +11,42 @@ import (
...
 func (bag MetricsBag) AddRawCounter(name string, value int64) {
 	bag["c:"+name] = value
 }
 
-// AddDeltaCounter adds a count of something that happened in the last second
+// AddDeltaCounter adds a metric of type counter to the bag and increments its value.
+// A counter is a cumulative metric that can only increase.
+// To be meaningful, `delta` must be positive.
```
arl, line 23: same as above

metrics.go
```diff
@@ -11,32 +11,42 @@ import (
...
-// AddHistogram adds a set of values to track their statistical distribution.
+// AddHistogram adds a metrics of type histogram to the bag and tracks its value.
+// A histogram samples observations and counts them in different 'buckets' in order
+// to track and show the statistical distribution of a set of values.
 func (bag MetricsBag) AddHistogram(name string, values []float64) {
 	bag["h:"+name] = values
 }
 
-// AddTimings adds a set of timings to track their statistical distribution.
+// AddTimings adds a metric of type time to the bag and traks its value.
```
arl, line 41: under the hood, there's no metric of  `time` or `timing`, this is just an wrapper around an histogram tracking time duration, such as latency
```suggestion
// AddTimings adds a metric of type histogram to the bag and tracks its value.
```

metrics.go
```diff
@@ -11,32 +11,42 @@ import (
...
+// Timing is basically a histogram but allows to sample values of type time.Duration.
+// A histogram samples observations and counts them in different 'buckets'.
 func (bag MetricsBag) AddTimings(name string, values []time.Duration) {
 	bag["t:"+name] = values
 }
 
-// Merge merges another MetricsBag into this 'bag'.
+// Merge merges metrics with the same name and types from another MetricsBag into this bag.
+// Counters and Deltas are summed up, Gauge is averaged, and Histograms and Timings concatenate.
```
arl, line 49:
```suggestion
// Counters are summed up, gauges are averaged, and histograms are concatenated.
```

arl (approved):
LGTM
This was the last pass, there are just some small stuff, you can merge once addressed, thanks!

PR #168, opened 2021-03-16
Remove flaky test from TCP input

arl:
#### :question: What

Fix some missing `Close` in `TCP` input and remove a flaky test. No need to fix it since this test (`TestTCPStopChunk`) only made sense if the closure semantics of the `TCP` input, once `TCP.Stop()` is called, was to wait for the currently bound connection, if any, to be closed by the client. But since this is not how  this input works, (i.e. it closes as soon as possible after receiving `Stop`) fixing the test to make it work would make it be the same as `TestTCPChunks`.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/input/tcp.go
+++ b/input/tcp.go
@@ -99,8 +99,11 @@ func (s *TCP) Run(inch chan<- *baker.Data) error {
 		ctxLog.WithFields(log.Fields{"addr": conn.RemoteAddr()}).Info("Connected")
 		wg.Add(1)
 		go func(conn *net.TCPConn) {
-			defer wg.Done()
-			s.handleStream(conn)
+			defer func() {
+				conn.Close()
+				wg.Done()
+			}()
+
 			err := s.handleStream(conn)
 			if err != nil {
 				ctxLog.WithError(err).WithFields(log.Fields{"error": err}).Error("Error when handling stream")
@@ -137,8 +140,6 @@ func (s *TCP) Stats() baker.InputStats {
 func (s *TCP) Stop() {
 	atomic.StoreInt64(&s.stop, 1)
 }
-
-	defer conn.Close()
 
 func (s *TCP) handleStream(conn *net.TCPConn) error {
 	// r, err := newFastGzReader(conn)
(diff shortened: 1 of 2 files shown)


input/tcp.go
```diff
@@ -85,21 +87,29 @@ func (s *TCP) Run(inch chan<- *baker.Data) error {
...
 		go func(conn *net.TCPConn) {
-			defer wg.Done()
-			s.handleStream(conn)
+			defer func() {
+				conn.Close()
+				wg.Done()
+			}()
+
+			if err := s.handleStream(conn); err != nil {
+				ctxLog.WithError(err).WithFields(log.Fields{"error": err}).Error("Error when handling stream")
```
tommyblue, line 111: lowercase?
```suggestion
				ctxLog.WithError(err).WithFields(log.Fields{"error": err}).Error("error when handling stream")
```

input/tcp.go
```diff
@@ -85,21 +87,29 @@ func (s *TCP) Run(inch chan<- *baker.Data) error {
 	}
 	defer l.Close()
 
+	wg := sync.WaitGroup{}
 	for atomic.LoadInt64(&s.stop) == 0 {
 		l.SetDeadline(time.Now().Add(1 * time.Second))
 		conn, err := l.AcceptTCP()
 		if err != nil {
-			if opErr, ok := err.(*net.OpError); ok && opErr.Timeout() {
+			if errors.Is(err, os.ErrDeadlineExceeded) {
```
tommyblue, line 95: asking: is this the same semantic as before?
I see `os.ErrDeadlineExceeded` represents a generic I/O timeout, but the doc for `opErr.Timeout()` is quite... well, empty 😅

tommyblue (approved):

input/tcp.go
arl: what must remain lowercases are error strings, since they can  be concatenated / wrapped within others. However here it's a log string, log strings should start with caps

input/tcp.go
arl: yes it's the same semantic as before. `opErr.Timeout` is very low level and was the only way when this input has been created, in our case it's preferred to use `os.ErrDeadlineExceeded` since it's the error we're looking for

PR #170, opened 2021-03-17
CI: coverage report in own workflow

arl:
#### :question: What

This PR is the result of having investigated if cross coverage can add some benefit to the project.
The answer is no:
 - first it's not a good practice in Go due to the essence of Go packages, a Go package encapsulates private behavior which no package one (apart possibly the one it's defined in) should rely upon. A package public API can and should be tested in the same package this API is defined.
 - it happens that code in Baker that doesn't appear in coverage report is not tested, even by other packages, so even cross coverage would not artificially increase the coverage number.
 
Given the previous reasons, I've just somewhat refined a bit the CI scripts with respect to code coverage:
 - coverage is now performed in its own Github action workflow, in parallel with the other workflows
 - the previous simplifies the Linux workflow
 - a further simplification of the Linux workflow, where all static checks are grouped in a single step, no need to multiply steps
 - coverage is only reported for pull-requests actions, no need to do them on pushes
 - simplified makefile by grouping related targets together
 - coverage report now ignores all test-helper packages such as `input/inputtest`, `output/outputtest`, etc.
 - fix a problem in coverage report where the commit SHA was not always retrieved

See coverage report for this PR [here](https://app.codecov.io/gh/AdRoll/baker/compare/170/tree)

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/.github/workflows/cover.yml
+++ b/.github/workflows/cover.yml
@@ -1,0 +1,33 @@
+name: Coverage
+on: [pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.x
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+            fetch-depth: 2
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Coverage
+        env:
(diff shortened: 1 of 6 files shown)


.github/workflows/cover.yml
```diff
@@ -0,0 +1,33 @@
+name: Coverage
+on: [pull_request]
+jobs:
+  build:
+    name: Build
```
lucarin91, line 5: Maybe it is better to call the job Coverage instead of Build
```suggestion
  coverage:
    name: Coverage
```

lucarin91 (approved):
LGTM, I just add a small naming suggestion

Makefile
```diff
@@ -3,24 +3,48 @@
...
+
+build-extra: ## Build an example baker binary with the "extra" components (sqlite3 output atm)
+	CGO_ENABLED=1 go build -v -tags cgo_sqlite -o baker-bin-example ./examples/advanced/
+
+#
+# Coverage reports
+#
+
+--ci-cover:
+	@go test -coverprofile coverage.txt -covermode atomic ./...
```
tommyblue, line 39: what is `@go`?

PR #171, opened 2021-03-18
fix: ExpandList behaviour if source is empty

lucarin91:
#### :question: What

The current implementation of the `ExpandList` filter has a strange behavior in the situation the input list is empty.
In particular the filter consider the empty list as a list with 1 empty elements, namely `""` -> `[""]`
 
For instance, consider the case in which the filter is configured to expand the list in the field `source` in such a way that `0 index` -> `f1` and `1 index` -> `f2`.

Thus, if we provide the following input record:
|f1|f2|source|
|--|---|---------|
|foo|bar||

The filter produces the corresponding output:
|f1|f2|source|
|--|---|---------|
||bar||

This PR changes the behavior of the filter and in the aforementioned situation it simply does nothing and forward. Thus, the output of the previous example will be:

|f1|f2|source|
|--|---|---------|
|foo|bar||

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TimestampRange filter accepts 'now' as range [#106](https://github.com/AdRoll/baker/pull/106)
 - Standardize the components' structs names [#105](https://github.com/AdRoll/baker/pull/105)
 - **Breaking** Change func FieldName to FieldNames (slice) as it allows to know the number of defined fields [#110](https://github.com/AdRoll/baker/pull/110)
-- ExpandList filter just forwards if the input list is empty. 
+- ExpandList filter just forwards if the input list is empty [#171](https://github.com/AdRoll/baker/pull/171)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TimestampRange filter accepts 'now' as range [#106](https://github.com/AdRoll/baker/pull/106)
 - Standardize the components' structs names [#105](https://github.com/AdRoll/baker/pull/105)
 - **Breaking** Change func FieldName to FieldNames (slice) as it allows to know the number of defined fields [#110](https://github.com/AdRoll/baker/pull/110)
-- ExpandList filter just forwards if the input list is empty [#171](https://github.com/AdRoll/baker/pull/171)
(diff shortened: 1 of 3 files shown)


filter/expand_list_test.go
```diff
@@ -8,8 +8,8 @@ import (
 
 func TestExpandList(t *testing.T) {
 	tests := []struct {
-		name string
-		list string
+		name   string
+		record string // default: ",,"
```
arl, line 12: Having a default for a test case is slightly confusing, IMO it's better for tests to be explicit, and so in this case setting `record` to `",,"` when this is the behavior of the case we're testing

filter/expand_list_test.go
```diff
@@ -19,20 +19,21 @@ func TestExpandList(t *testing.T) {
 		wantErr bool
 	}{
 		{
-			name: "extract 2 value",
-			list: "value1;value2",
+			name:   "extract 2 value",
```
arl, line 22:
```suggestion
			name:   "extract 2 values",
```

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TimestampRange filter accepts 'now' as range [#106](https://github.com/AdRoll/baker/pull/106)
 - Standardize the components' structs names [#105](https://github.com/AdRoll/baker/pull/105)
 - **Breaking** Change func FieldName to FieldNames (slice) as it allows to know the number of defined fields [#110](https://github.com/AdRoll/baker/pull/110)
+- ExpandList filter just forwards if the input list is empty [#171](https://github.com/AdRoll/baker/pull/171)
```
arl, line 58:
```suggestion
- ExpandList filter just forwards if the source field is empty [#171](https://github.com/AdRoll/baker/pull/171)
```

filter/expand_list_test.go
```diff
@@ -136,6 +150,9 @@ func TestExpandList(t *testing.T) {
 			if tt.source == "" {
 				tt.source = "source"
 			}
+			if tt.record == "" {
+				tt.record = ",,"
+			}
```
arl, line 155: see comment above

filter/expand_list.go
```diff
@@ -119,6 +119,11 @@ func (f *ExpandList) Process(l baker.Record, next func(baker.Record)) {
 	atomic.AddInt64(&f.numProcessedLines, 1)
 
 	list := l.Get(f.source)
+	if len(list) == 0 {
+		next(l)
```
arl, line 123: One can understand why we early return here only if they know the behavior of `bytes.Split`, I'd add a small comment to help future readers here

arl (approved):
Good catch 👍🏽 

Nothing to add on the code itself, but please see comments before merging

PR #174, opened 2021-03-25
Clone configs instead of reusing pointer in desc

arl:
To trigger the problem fixed by this change you have to:
 - use a component with a Config structure using a map
 - somewhat manage to create, in the same process, at least 2 baker
 topologies using this component, each of which having different keys in
 the map.

Even if not impossible, it would be rather improbable to have this in a
normal production usage. However it's really easy to trigger with tests!

`go test -run foo` does exactly that`. It launches X times the tests
containing foo in their name, all in the same process.

The problem arises when copying the Desc.Config (an interface{}) to the
DecodedConfig field of the final component instance (also an
interface{}). The DecodedConfig is then used as destination for TOML
decoding.  But the thing is that the TOML library we're using merges
maps in TOML with the possibly existing content of the destination map
(which is probably a feature, really!). Since we've assigned an
interface onto another, we've actually just made the pointer part of the
destination interface (an interface is, under the hood, 2 fields, a type
and an interface) point to the pointer in the source interface.

In other words, we have decoded our TOML in the destination interface
DecodedConfig, but also in Desc.Config.

To solve this, it's enough to call cloneConfig in other words to create
new instances of the Config structs.

#### :question: What

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Do not call Output.Stats() twice in `StatsDumper` [#173](https://github.com/AdRoll/baker/pull/173)
+- Call cloneConfig instead of resuing pointer in Desc configurations [#174](https://github.com/AdRoll/baker/pull/174)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


tommyblue (approved):
good catch!

PR #175, opened 2021-03-29
Add the Truncate filter

tommyblue:
#### :question: What

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


filter/slice.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+
+	"github.com/AdRoll/baker"
+)
+
+var SliceDesc = baker.FilterDesc{
+	Name:   "Slice",
+	New:    NewSlice,
+	Config: &SliceConfig{},
+	Help:   "Slices the source field value to the given bytes length and start index and saves the value to the destination field. If the start index is greater than the field value lenght, set an empty string to destination",
```
arl, line 13: Both in python and in Go, slice expression is made of `[low:high]` where:
 - low and high are both indices
 - both can be omitted, which result in low being 0 and high being the index of the last element

There's no parameter that directly represent the length and IMHO we should follow this philosophy with the Slice filter.

filter/slice.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	dst, ok := cfg.FieldByName(dcfg.Dst)
+	if !ok {
+		return nil, fmt.Errorf("cannot find destination field \"%s\"", dcfg.Dst)
+	}
+
+	if dcfg.Length < 1 {
+		return nil, fmt.Errorf("invalid length %d", dcfg.Length)
+	}
```
arl, line 47: with `low:high` you'd have obviously to check that low <= high

filter/slice.go
```diff
@@ -0,0 +1,82 @@
...
+
+	end := f.startIdx + f.length
+	if end > l {
+		end = l
+	}
+
+	r.Set(f.dst, v[f.startIdx:end])
+	next(r)
+}
```
arl, line 77: I think we should always allocate a new a slice here. Slices are just views of their backing array and multiple slices can share the same backing array. That's what happens if the code takes the path at l75, 
```go
r.Set(f.dst, v[f.startIdx:end])
```
If one of these fields happens to be modified in-place, then the other will also be modified.

While we're at it, if we always create a new slice, can we try to modify the method so that there's only one single call to `next`?

arl (changes requested):
- i think we should stick with low:high type of slicing, rather than low:length (see comment)
- some typos: `s/lenght/length`
- potential problem in process

filter/slice.go
tommyblue: oh, damn. right

filter/slice.go
tommyblue: ok

filter/slice_test.go
```diff
@@ -0,0 +1,174 @@
...
+		src         string
+		dst         string
+		start       int
+		end         int
+		record      []byte
+		want        []byte
+		wantInitErr bool
+		wantErr     bool
+	}{
+		{
```
arl, line 33: Maybe we could add a case where both start and end are defaults, which should always copies the whole field into the destination
And another, a config error, where start > end

filter/slice.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+
+	"github.com/AdRoll/baker"
+)
+
+var SliceDesc = baker.FilterDesc{
+	Name:   "Slice",
+	New:    NewSlice,
+	Config: &SliceConfig{},
+	Help:   "Slices the source field value using start/end indexes and saves the value to the destination field. If the start index is greater than the field value length, set an empty string to destination",
```
arl, line 13: nit: Maybe _copies_ instead of _saves_?

arl (approved):
LGTM!
one nit comment about the wording and some possibly missing test cases

filter/slice_test.go
tommyblue: ok

tommyblue:
@arl check again please as I also added a small optimization: no need to check this as it's already covered by previous checks

```go
} else if sz > len(src) {
	sz = len(src)
}
```

filter/slice.go
```diff
@@ -0,0 +1,86 @@
...
+
+	sl := make([]byte, sz)
+
+	if f.startIdx < len(src) {
+		copy(sl, src[f.startIdx:end])
+	}
+
+	r.Set(f.dst, sl)
+	next(r)
+}
```
arl, line 81: it seems slightly too complex to me 😛 
I might have forgotten something so take it with a grain of salt but I think we need 3 checks

This makes sure we'll get an empty field in this case
```go
	if start > len {
		start = 0
	}
```

This ensures that we take the whole field if end is not given
 ```go
	if end == 0 {
		end = len
	}
```

Finally this ensures we're not going past the len in any case
```go
	if end > len {
		end = len
	}
```

Once all these checks are done, you can use this trick to create a copy of the field without actually using `copy` and so without having to compute `end-start`.
`append` on a nil slice creates a new copy.
```go
	var empty []byte
	r.Set(dst, append(empty, s[start:end]...))
```

arl (dismissed):
possible simplification in `Process`

filter/slice.go
tommyblue: thanks for the hint!
I refactored the checks

filter/slice.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+
+	"github.com/AdRoll/baker"
+)
+
+var SliceDesc = baker.FilterDesc{
+	Name:   "Slice",
+	New:    NewSlice,
+	Config: &SliceConfig{},
+	Help:   "Slices the source field value using start/end indexes and copies the value to the destination field. If the start index is greater than the field value length, set an empty string to destination",
```
arl, line 13: We should also document what's happening when the end index is greater than the field length

filter/slice.go
```diff
@@ -0,0 +1,84 @@
...
+	endIdx   int
+}
+
+// NewSlice creates a new Slice filter
+func NewSlice(cfg baker.FilterParams) (baker.Filter, error) {
+	dcfg := cfg.DecodedConfig.(*SliceConfig)
+
+	src, ok := cfg.FieldByName(dcfg.Src)
+	if !ok {
+		return nil, fmt.Errorf("cannot find source field \"%s\"", dcfg.Src)
```
arl, line 37: nit: we can use `%q` (quoted string) rather than `\"%s\"`

filter/slice.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	dst, ok := cfg.FieldByName(dcfg.Dst)
+	if !ok {
+		return nil, fmt.Errorf("cannot find destination field \"%s\"", dcfg.Dst)
```
arl, line 42: same

arl (approved):
Last review for me. LGTM! Only some doc nits

arl:
@tommyblue I addressed my own comments since we want to bump baker today and include this filter. There were only documentation improvements and nitpicks. I also added a note in the component help, stating the indexes are 0-based and encoding-free since they act on bytes

PR #173, opened 2021-03-22
test: update StatsDumper integration test

lucarin91:
#### :question: What

Updates the integration tests of the `StatsDumper` checking that the gathered metrics are correctly published to the `MetricsClient`.

In addition to that, the PR adds the following changes/fixes to the code base:

- publish to the `MetricsClient` custom histogram and custom timing metrics gathered through the `MetricsBag`
- remove the filters `NumProcessedLines` counter, which is never used
- fix a bug that calls two times the `Stats` function of the Outputs
- set as public the `metrics` field of the `Topology` structs

<br/>

**TODO** update the following website pages:
- [ ] https://getbaker.io/docs/how-tos/metrics/
- [ ] https://getbaker.io/docs/how-tos/create_filter/

#### :white_check_mark: Checklists

- [x] Have the changes in this PR been functionally tested?
- [ ] Have new components been added to the documentation website?
- [x] Has `make gofmt-write` been run on the code?
- [x] Has `make govet` been run on the code? Has the code been fixed accordingly to the output?
- [x] Have the changes been added to the [CHANGELOG.md](./CHANGELOG.md) file?
- [x] Have the steps in [CONTRIBUTING.md](./CONTRIBUTING.md) been followed to update a Go module?

PR #176, opened 2021-03-30
remove NumProcessedLines in filter doc

lucarin91:
#### :question: What

Update the documentation according to PR #173.
The references of the filter `NumProcessedLines` counter are removed.


arl (approved):
LGTM!

PR #178, opened 2021-03-31
input: kcl: add error to checkpointing error log

arl:
When there's a checkpointing error, we were not logging the cause.


lucarin91 (approved):
LGTM

PR #179, opened 2021-04-06
fix: force go get of the baker main branch

lucarin91:
#### :question: What

Small fix on `gen-components` task: use `@main` instead of `@latest` in go get. Indeed, the `@latest` tag gets the latest release of the package instead of the latest version of the `main` branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ git-update:
 setup: setup-git ## Setup Hugo locally 
 	npm install postcss-cli autoprefixer postcss
 
-gen-components: ## Generate components markdown from baker@latest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ setup: setup-git ## Setup Hugo locally
 	npm install postcss-cli autoprefixer postcss
 
 gen-components: ## Generate components markdown from baker@latest
```
Oppen, line 20: What about the comment?

arl: yeah. good point. @lucarin91 once you fix the comment we can merge this

arl (approved):
LGTM

PR #181, opened 2021-04-06
Bump gozstd gokcl

arl:
#### :question: What

 * Bump vmware-go-kcl to 022ec8d (latest version on  `master`, should contain a number of fixes/improvements)
 * Bump gozstd to v1.9.0 (latest tagged version, wraps zstd 1.4.7, which contains a number of fixes and perf improvements)

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


.github/workflows/linux.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           make gofmt
           make govet
           go mod verify
-          go install golang.org/x/tools/cmd/goimports
+          go install golang.org/x/tools/cmd/goimports@latest
```
lucarin91, line 31: Just out of curiosity, why you need the `@latest` tag instead of the default `@upgrade`? [version-queries](https://golang.org/ref/mod#version-queries)

lucarin91 (approved):
LGTM

.github/workflows/linux.yml
arl: `go install` doesn't default to @upgrade, it's `go get` that does.

Also, from the docs you linked:
> Since Go 1.16, go install is the recommended command for building and installing programs. When used with a version suffix (like @latest or @v1.4.6), go install builds packages in module-aware mode, ignoring the go.mod file in the current directory or any parent directory, if there is one.
>
> go get is more focused on managing requirements in go.mod. The -d flag is deprecated, and in a future release, it will always be enabled.

So, since we know use go 1.16, just doing go install (without latest) will modify the current go.mod since we're running go install from a repository containing a `go.mod`.
However by adding `@latest` we're just installing the latest version of goimports _in the system_, but this is not anymore related to the fact that we're in a repository with a go.mod or not.

Tips I use to avoid any confusion:
 - use go get to do go.mod related actions
 - use go install for system related actions
 - and use a specific version query, always, for both, whatever `@va.b.c` `@latest` `@upgrade` , etc

lucarin91: Thanks for the clear explanation :smiley:. I am wondering if it will be better to use `go install` instead of `go get` in the [documentation generation](https://github.com/AdRoll/baker/blob/gh-pages/Makefile#L21). I recently open PR #179 to force the download of the last baker version in the `main` branch. In this way, we could also completely remove the `go.mod` file in the scripts folder, but this is a conversation for another PR.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove `NumProcessedLines` metric from FilterStats [#173](https://github.com/AdRoll/baker/pull/173)
 - Add the Crypt filter [#177](https://github.com/AdRoll/baker/pull/177)
+- Bump vmware-go-kvl and gozstd [#181](https://github.com/AdRoll/baker/pull/181)
```
tommyblue, line 18:
```suggestion
- Bump vmware-go-kcl and gozstd [#181](https://github.com/AdRoll/baker/pull/181)
```

tommyblue (approved):

PR #186, opened 2021-08-09
Fix/filewriter race

arl:
#### :question: What
This rewrites a big part of the FileWriter output to make it more resilient, adding many tests and fixes at least 2 race conditions (from which an old one allowing us to close BAK-40 ticket). Overall, the output now waits for all the workers to be terminated before exiting from `FileWriter.Run`, which was not the case before, leading to some trailing garbage in produced files, since not all bytes have had the time to be written.

For long running topologies, the problem was mitigated by the quantity of records processed, however, for short-running topologies (or with a single file as BAK-40), often the file contained garbage.

 - add zip_agnostic package (useful to testing)
 - improve doc
 - check template at configuration  time (drop panic)
 - etc.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


output/filewriter_test.go
```diff
@@ -62,3 +80,398 @@ func TestFileWriterConfig(t *testing.T) {
 		})
 	}
 }
+
+// testFileWriterCompareInOut sends numRecords records and wait for the given
+// duration between each send, to a FileWriter where PathString is set to the
+// concatenation of the path components.
+func testFileWriterCompareInOut(numRecords int, wait, rotate time.Duration, comps ...string) func(*testing.T) {
```
lucarin91, line 87: we can probably add another check to verify that all the created files are exactly the same sent to the upload channel. It will make sure that no other files are created during the output execution.

lucarin91, line 87: we can probably add another check to verify that all the created files are exactly the same sent to the upload channel. It will make sure that no other files are created during the output execution.

output/filewriter_test.go
```diff
@@ -62,3 +80,398 @@ func TestFileWriterConfig(t *testing.T) {
...
+	const procs = 1
+	const rotate = 1 * time.Microsecond
+	testFileWriterIntegrationCheckRecords(t, filepath.Join("TMPDIR", "{{.Rotation}}-out.csv.zst"), procs, rotate)
+}
+
+func TestFileWriterIntegrationUUID(t *testing.T) {
+	const procs = 8
+	const rotate = -1
+	testFileWriterIntegrationCheckRecords(t, filepath.Join("TMPDIR", "{{.UUID}}-out.csv.zst"), procs, rotate)
+}
```
lucarin91, line 413: Are this tests in separated function because the `testFileWriterIntegrationDeterministic` uses `t.Names()` to access the correct golden file? If not probably they are more clear in a single function with multiple `t.run(`.

pkg/zip_agnostic/reader_test.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, fname := range fnames {
+		t.Run(fname, func(t *testing.T) {
+			f, err := os.Open(fname)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			r, err := NewReader(f)
+			if err != nil {
+				t.Fatalf("zagnosticReader returns %v", err)
```
lucarin91, line 34: is this a typo?
```suggestion
				t.Fatalf("ZipAgnosticReader returns %v", err)
```

output/filewriter_test.go
```diff
@@ -62,3 +80,398 @@ func TestFileWriterConfig(t *testing.T) {
...
+func TestFileWriterIntegrationRotationIndex(t *testing.T) {
+	testFileWriterIntegrationDeterministic(t, filepath.Join("TMPDIR", "{{.Rotation}}", "out.csv.zst"))
+}
+
+func TestFileWriterIntegrationTimestamp(t *testing.T) {
+	const procs = 1
+	const rotate = -1
+	testFileWriterIntegrationCheckRecords(t, filepath.Join("TMPDIR", "{{.Year}}{{.Month}}{{.Day}}{{.Hour}}{{.Minute}}{{.Second}}-out.csv.zst"), procs, rotate)
+}
+func TestFileWriterIntegrationProcs(t *testing.T) {
```
lucarin91, line 397: Here is missing a test in which both `index` and `field0` are used. This is the situation in which we have the greater number of concurrent workers `nproc` FileWriter times a fileWorker for each `field0` occurrence. Maybe is a case worth testing.

lucarin91 (approved):
great improvement! I add some comments mostly about the style.

output/filewriter.go
```diff
@@ -140,71 +207,168 @@ func (cfg *FileWriterConfig) fillDefaults() {
 // fileWorker manages writes to a file and its periodic rotation.
 type fileWorker struct {
 	in   chan []byte
-	done chan bool
-	upch chan<- string
+	done chan struct{}
 
 	cfg *FileWriterConfig
```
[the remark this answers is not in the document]
arl: ha yes, forget to do that; good catch

output/filewriter.go
```diff
@@ -219,106 +383,36 @@ func (fw *fileWorker) makePath() string {
 		"Field0":   fw.replFieldValue,
 	}
 
-	err := fw.pathTemplate.Execute(&doc, replacementVars)
+	err := tmpl.Execute(&buf, replacementVars)
 	if err != nil {
 		panic(err.Error())
```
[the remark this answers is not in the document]
arl: Since the template had already been parsed, en error at `Execute` is a developer error and should be caught asap so i think a panic is right here since a failure can't originate from the user.

However there's an future improvement we could with `PathString` templattng. We could (and we should) at configuration time, call `Execute` once and verify that the corresponding path is valid. For now, if I for example set `pathstring = "/{{.Hello}}"` in the TOML, the generated path is invalid but only sees it when `newWorker` is called

output/filewriter_test.go
```diff
@@ -35,7 +53,7 @@ func TestFileWriterConfig(t *testing.T) {
 		// error cases
 		{
```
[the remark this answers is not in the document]
arl: yes, good point. Also I think I should fix the case I'm talking about in the previous comment:
""{{.Hello}}"  is a valid template, so not caught at config time, but invalid for our use case since, once executed this gives the `/<no value>` string, which is not a valid path.

output/filewriter_test.go
arl: I tend to use test tables for tests having a clear input and output. 
And yes, deterministic tests uses t.Names while the others all use the same golden file.

pkg/zip_agnostic/reader_test.go
arl: good catch, it was the previous name :-)

output/filewriter_test.go
arl: 👍🏽

output/filewriter_test.go
arl: done

output/filewriter_test.go
arl: done

output/filewriter_test.go
arl: done

output/filewriter_test.go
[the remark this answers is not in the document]
arl: I've added a test case with an malformed pattern.
However, i have a bit of thought to trying to assert that the generated path will be valid, but it's futile. Valid path is an OS-dependent thing. Moreover, if `{{.Field0}}` is in the string, then we can't possibly know that generated path are always going to be valid. 
I've improved the help string for Field0 with these remarks, plus one about the cardinality if the field. In other words,  is the user uses Field0, with great powers comes great responsibility.

lucarin91 (approved):
:+1:

PR #183, opened 2021-07-21
Fix/filewriter bug

arl:
#### :question: What

 - Improve Filewriter doc about `{{.Field0}}` in template.
  - fix bug: if `{{.Field0}}` is provided, at least one field must be present in `[output.fields]`. Apart from the first field, providing extra fields in `[output.fields]` is useless since `FileWriter` only uses the first anyway. This output is a raw output, that is, it writes serialized records to file, no just sets of fields.
  - add test
 
#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

#### :white_check_mark: Checklists

_This section contains a list of checklists for common uses, please delete the checklists that are useless for your current use case (or add another checklist if your use case isn't covered yet)._

- [x] Is there unit/integration test coverage for all new and/or changed functionality added in this PR?
- [x] Have the changes in this PR been functionally tested?
- [ ] Have new components been added to the related `all.go` files?
- [ ] Have new components been added to the documentation website?
- [x] Has `make gofmt-write` been run on the code?
- [x] Has `make govet` been run on the code? Has the code been fixed accordingly to the output?
- [x] Have the changes been added to the [CHANGELOG.md](./CHANGELOG.md) file?
- [x] Have the steps in [CONTRIBUTING.md](./CONTRIBUTING.md) been followed to update a Go module?

PR #187, opened 2021-08-09
feat: add ExternalMatch filter

lucarin91:
#### :question: What

Add the `ExternalMatch` filter. 

~:warning: This PR requires the `zip_agnostic` pkg that will be introduced with PR #183.~


testutil/helpers.go
```diff
@@ -97,3 +98,8 @@ func LessLogging() (reset func()) {
 	logrus.SetLevel(logrus.ErrorLevel)
 	return func() { logrus.SetLevel(lvl) }
 }
+
+// PathToUri converts a file path to an URI. Compatible with both Windows and Linux paths.
+func PathToURI(p string) string {
+	return "file://" + filepath.ToSlash(p)
+}
```
arl, line 105: This function is only used twice in `external_match_test.go` . Since it's really one-line function, I think it's preferable to:
 - not use it at all, and use `file://` + filepath.ToSlash(p)` where we need.
 - if you prefer a function, you can create it, non exported, in `external_match_test.go`

In `list_test.go` this is used many times so the function is useful there. 

_A little copying is better than a little dependency._ https://go-proverbs.github.io/

Moreover, after that, the diff of the PR will get just get back to the `ExternalMatch` filter and its tests

arl (changes requested):
adding the helper function to testutil adds dependency where IMHO it was not needed; see comment

PR #188, opened 2021-08-09
fix non-64-bit aligmment of atomic variables

arl:
On 32-bits platforms, atomically accessing a 64-bit variable
which is not 64-bit aligned panics, since these accesses are
not guaranteed to effectively be performed atomically.

hint: in Go, a global variable, as well as the first field of
a struct, is guaranteed to be 64-bit aligned, even on 32-bit
platforms.


Oppen (commented):
LGTM

lucarin91 (approved):
:+1:

PR #196, opened 2021-11-15
Input/sqs/load from multiple buckets

arl:
#### :question: What

This PR adds some integration tests for the SQS input and verifies that the S3 files to download might be located on multiple buckets.
Also, this clarifies the role of the `Bucket` configuration parameter. The help string for SQS was misleading about it and it seemed the bucket was a way to restrict the S3 path received via SQS. However the `Bucket` parameter is only used in a very specific case: if the incoming messages do not contain `s3://bucket`. In this case the downloaded path is the path which is the concatenation of `s3://BUCKET/path`.

note: the only way to restrict downloading to a single (or multiple) S3 bucket(s) is to do so with the `FilePathFilter` regular expression.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


lucarin91 (approved):
LGTM

PR #201, opened 2022-04-19
Add count and tag filter

arl:
#### :question: What

Add a new filter, called `CountAndTag`. This filter publishes a metric that simply counts all records that pass through, and breaks them down, with metrics tags, by the value of a configured field.
Records having an empty string as `Field` value are counted nonetheless, using the value configured as `DefaultValue` as tag value

note: performance has been measured on a real-life pipeline, with datadog as metrics client implementation, and both cpu and memory were dominated by io (reading from input, writing to output) so we consider the performance as acceptable, in fact it greatly depends on the metrics client implementation itself, since there will be one call per record, per filter chain process. That's not a problem per se since metrics libraries are optimized for this exact use case.

### Example visualization for a topology configured as follows:
```toml
[[filter]]
name = "CountAndTag"
[filter.config]
DefaultValue = "none"
Field = "currency"
Metric = "currency_name"
```

image

While at it, fix a flaky windows test of the `ExternalMatch` filter.

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.

### DIFF
--- a/filter/count_and_tag.go
+++ b/filter/count_and_tag.go
@@ -7,12 +7,10 @@ import (
 	"github.com/AdRoll/baker"
 )
 
-const countAndTagHelp = `Publish a metric that simply counts all records that pass through, using one of the field as value for metric tag.
+const countAndTagHelp = `Publish a metric that simply counts all records that pass through, and breaks them down, with metrics tags, by the value of a configured field.
+Records having an empty string as Field value are counted nonetheless, using the value configured as DefaultValue as tag value.
 
-By default, records having where the chosen field is empty are counted, but not tagged.
-You can change that behavior by setting a default tag value for these records, with the DefaultTagValue configuration parameter.
-
-NOTE: a special attention should be given in the selected Field used to tag records.
+NOTE: a special attention should be given in the selection of the Field used to tag records.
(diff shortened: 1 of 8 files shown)


filter/count_and_tag.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return f, nil
+}
+
+func (f *CountAndTag) Stats() baker.FilterStats {
+	return baker.FilterStats{}
+}
+
+func (f *CountAndTag) Process(l baker.Record, next func(baker.Record)) {
+	v := l.Get(f.fidx)
```
lucarin91, line 66: I think we need to check if `v` is a valid tag value, otherwise, we will silently fail. For instance, here are the tag specification of Datadog https://docs.datadoghq.com/getting_started/tagging/#defining-tags

lucarin91 (approved):

filter/count_and_tag.go
arl: There's not much we can do about that. Tag validity depends on the metrics client used, what's valid for datadog might not be valid for another client. But anyways, metrics, are like logs observability features that are best effort, but should not in any way have consequences on the pipeline outcome. 
There's a warning in the component doc about being careful about tag value cardinality, what i can do is adding another warning about value validity.

PR #203, opened 2022-04-29
Filewriter/rotatesize option

arl:
#### :question: What

Add `RotateSize` option to `Filewriter` output.
`RotateSize` is the file size which when reached triggers a file rotation. Can be cumulated with `RotateInterval`. 0 to disable. 

The `RotateSize` configuration parameter is a new type of component parameter representing a size. Its value can be provided:
 - as an integer number of bytes,
 - as a string with an SI unit
 - as a string with IEC unit
 
Examples: `12000`, `"12KB"`, `"1MB"`, `"1MiB"`, etc." See [here for more units](https://en.wikipedia.org/wiki/Binary_prefix)

#### :hammer: How to test

1. List all steps necessary;
2. To test this pull request.


output/filewriter.go
```diff
@@ -352,17 +333,65 @@ func newWorker(cfg *FileWriterConfig, tmpl *template.Template, replFieldValue st
...
+			WindowLog:        fw.cfg.ZstdWindowLog,
+		})
+		wc = makeWriteCloser(zstdw, zstdw.Close)
+	} else {
+		// Only way to for gzip.NewWriterLevel to fail is to pass an
+		// incorrect compression level.
+		wc, _ = gzip.NewWriterLevel(countw, gzip.BestSpeed)
+	}
+
+	// Close the writers (the correct is important here):
```
lucarin91, line 377: Probably a typo
```suggestion
	// Close the writers (the order is important here):
```

output/filewriter.go
```diff
@@ -352,17 +333,65 @@ func newWorker(cfg *FileWriterConfig, tmpl *template.Template, replFieldValue st
...
 }
 
+func (fw *fileWorker) newFile(path string) (io.WriteCloser, func() int64, error) {
+	f, err := os.Create(path)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	bufw := bufio.NewWriterSize(f, fileWorkerChunkBuffer)
+	countw := newCountingWriter(bufw)
```
lucarin91, line 362: If I understand correctly the code, here we are counting the not compressed bytes, so the final file size could have a lower size than the one selected by the user. This could be the intended behavior, but maybe we could make it clear on the documentation.

lucarin91 (approved):
LGTM, I add some small comments, nothing major.

output/filewriter.go
arl: We're counting the compressed bytes here (i.e the final size of the compressed file). 
The `countingWriter` wraps the `bufio.Writer`, which itselfs wraps (and buffers) the final file. The bytes written into `countw` comes from the zip writer.
The test confirms that -> https://github.com/AdRoll/baker/blob/8271279a50346a20bf2ec98d0ee6f4838983f728/output/filewriter_test.go#L604-L614

Let me know if that's clear or if there's some misunderstanding.

lucarin91: You are right I misread the code. Thanks for the clarification.

PR #210, opened 2022-09-28
Implementing a new config QueueNames to sqs Input.

danimaribeiro:
Queue prefixes could lead to undesired behaviour, this new option makes the use strict to specific queues

#### :question: What

Implementing a new config QueueNames to sqs Input. 
When using deadletter queues in SQS we often do not expect Baker to read from that queue, but because the default is to add the suffix to the queue name to indicate is a dead-letter queue, Baker ends up reading from that queue too.

With this new parameter we are strict to which queues we want to read from without surprises if some queue with similar name is created.

#### :hammer: How to test

Change the QueuePrefixes parameter to QueueNames - it should behave the same way, but it will listen only to that specific queue.

- [X] Is there unit/integration test coverage for all new and/or changed functionality added in this PR?
- [X] Have the changes in this PR been functionally tested?
- [X] Have new components been added to the related `all.go` files?
- [X] Have new components been added to the documentation website?
- [X] Has `make gofmt-write` been run on the code?
- [X] Has `make govet` been run on the code? Has the code been fixed accordingly to the output?
- [X] Have the changes been added to the [CHANGELOG.md](./CHANGELOG.md) file?
- [X] Have the steps in [CONTRIBUTING.md](./CONTRIBUTING.md) been followed to update a Go module?


arl (approved):
LGTM 👍🏽 

Thank you!

